PR #3, opened 2018-10-08
Slower bwtest graph

mwfarb:
Allow 60 sec of data in graph, not 20 sec.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/3)


mwfarb:
@perrig Suggested slowing down the graph such that 1 minute of history could be viewed.

PR #7, opened 2018-11-29
[WIP] adding 'bat' application and restructuring repository to include a 'lib' folder

chaehni:
This PR adds the 'bat' application to scion-apps.
Additionally, a new 'lib' folder is created in which commonly used scion library code will be gathered.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/7)


chaehni:
To do:
* move other library code to lib and update applications to make use of it
* make applications use their own vendor folder
* licensing of bat application

lib/shttp/server.go
```diff
@@ -0,0 +1,125 @@
...
+		return initErr
+	}
+
+	network := snet.DefNetwork
+	sconn, err := network.ListenSCION("udp4", laddr)
+	if err != nil {
+		return err
+	}
+
+	return ServeSCION(sconn, handler, certFile, keyFile)
```
matzf: Why not just create a `Server` object here and call `Server::ListenAndServe` (just like in `ServeSCION` below)?

PR #36, opened 2019-01-23
Port webapp to scionlab jan2019

mwfarb:
- Fixed compile errors.
- Fixed services enumeration on AS Topology tab.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/36)

PR #37, opened 2019-01-24
Adding asynchronous health tests feedback

### LINKED ISSUE #19: [webapp] Use asynchronous health check feedback
Currently, the health check runs as a single block. When future scripts run over a few seconds, users will wonder what's happening without better feedback. This would add:
- Max 60s timeout for any single bash script.
- UI label "In Progress" in amber for in progress scripts.
- UI label "Pending" in grey for scripts prior to execution.
- Health check landing page will show all scripts as "Pending" on page load.
- Background tasks to run scripts and update page.

mwfarb:
- Moved db general methods to db.go.
- Added elapsed time to each test.
- Changed beacon test to find last verification within 10s.
- Changed beacon test to timeout after 20s.
- Changed all tests to timeout after 60s.
- Added troubleshooting add tutorials links.
- Modified results to allow viewing of test result after each completes.
- Closes #19.

For testing: try starting and stopping the infrastructure so that beacons will stop arriving. Then, reload the Health page, after 10s, the test will begin checking for beacons on a 1s loop for 20s.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/37)

### DIFF
--- a/webapp/lib/health.go
+++ b/webapp/lib/health.go
@@ -10,26 +10,42 @@ import (
     "os/exec"
     "path"
     "path/filepath"
-    "strconv"
     "strings"
+    "time"
 )
 
-// Tests holds the JSON array for all health checks.
-type Tests struct {
-    Tests []Check `json:"tests"`
+var defFileHealthCheck = "tests/health/default.json"
+var resFileHealthCheck = "data/healthcheck-result.json"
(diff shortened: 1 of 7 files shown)


mwfarb:
After scionlab_nextversion, depends on #36.

PR #46, opened 2019-03-04
[pathdb_dump] Fix -t option timestamp

mwfarb:
The database seems to store one timestamp in seconds and the other in nanoseconds.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/46)


mwfarb:
**Before:** Expires in: 431033h32m26.373716166s
**After:** Expires in: -5h59m20.337142249s

PR #60, opened 2019-04-09
Fixed build and added missing vendor entries.

### LINKED ISSUE #57: govendor list +o returns non-empty
When running `govendor list +o` on current master, it returns some missing libraries:
```
  m github.com/jtolds/gls                            
  m github.com/smartystreets/assertions              
  m golang.org/x/image/font                          
  m golang.org/x/image/font/basicfont                
  m golang.org/x/image/math/fixed                    
  m roughtime.googlesource.com/go/client/monotime    
  m roughtime.googlesource.com/go/config             
  m roughtime.googlesource.com/go/protocol
```
Some of them come from our `roughtime` app, but I don't know if all of them do. 
Find out why we have these missing libraries and modify consequently our `vendor.json`

juagargi:
I had to add a different origin for the three roughtime.googlesource.com libraries, as
I could not make it work with govendor fetch or go get for the life of me.
Added roughtime to the makefile and fixed a missing call in the time server to use our scion util library.

The second commit comes from a patch from @mwfarb to add missing dependencies from `webapp`.

And while I was at it, I removed other unused entries so now `govendor list +o` returns empy.

Closes #57

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/60)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,6 @@ sensorapp/sensorfetcher/sensorfetcher
 sensorapp/sensorserver/sensorserver
 webapp/webapp
 tools/pathdb_dump/pathdb_dump
+netcat/netcat
+roughtime/timeclient/timeclient
+roughtime/timeserver/timeserver
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,7 @@
 .PHONY: all clean
 
 ROOT_DIR=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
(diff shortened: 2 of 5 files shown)


mwfarb:
I think `golang.org/x/image/font/plan9font` got pulled into my patch when I did `govendor add +e`. Since `webapp/tests/imgtest/imgserver/local-image.go` compiles without it, it can be removed.

PR #65, opened 2019-04-26
Initial commit for scion-ssh

N2D4:
Feature-complete for initial version, but I'll try to document and clean-up a little more before the merge

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/65)


N2D4:
Note that I'm using kingpin for argument parsing here because `flags` isn't really compatible with the argument syntax used by Unix SSH, but I'm open for alternatives

PR #67, opened 2019-05-15
ignore netcat build

mwfarb:
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/67)


mwfarb:
Oops. I see this was added in #60.

PR #80, opened 2019-06-21
[bwtester] Fix panic on uninitialized network

chaehni:
Fixes a panic that occurs when bwtestclient tries to fetch paths from an uninitialized network

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/80)


matzf:
This is no longer applicable (the `appnet` package initializes the default network and panics in case of an error).

PR #81, opened 2019-06-21
[webapp] Add Multi-Path Colors & bug fixes

### LINKED ISSUE #69: [webapp] handle color logging in bwtester path match
Bwtesterclient will output paths in the console. Webapp translates it's GUI paths to the same console format to choose a path in console interactive mode. The colorizing of paths in bwtesterclient is being interpreted by webapp to provide 0 path matches. 

We need to modify webapp to handle this.

mwfarb:
- Adds deterministic colors for each unique path.
- Adds path colors to path selection list.
- Adds/changes path colors to bwtest and echo graph data points.
- Adds path # hops as a badge next to path selection list.
- Adds gray color for graph data points without matching paths.
- Changes to faint-red color for graph error data points, not red.
- Fixes bwtest failure to match interactive paths due to ANSI escapes.
- Fixes bwtest failure to match default path.
- Improves regex cases to record echo errors.
- Adds stderr to command parsing stream with stdout.
- Fixes overlapping continuous switch conditions between user-server.
- Fixes #69.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/81)

### DIFF
--- a/webapp/lib/bwcont.go
+++ b/webapp/lib/bwcont.go
@@ -34,6 +34,7 @@ var reErr3 = `(?i:error:\s*)([\s\S]*)`
 var reErr4 = `(?i:eror:\s*)([\s\S]*)`
 var reErr5 = `(?i:crit:\s*)([\s\S]*)`
 var reUPath = `(?i:using path:)`
+var reSPath = `(?i:path=*)"(.*?)"`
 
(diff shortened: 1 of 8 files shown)


mwfarb:
Screenshot from 2019-06-21 15-47-12

PR #86, opened 2019-07-10
Allow building from an almost empty system

juagargi:
It requires `go` and some other standard packages such as e.g. `sudo`.
Fixes issues with the incorrect `vendor.json` file, as well as not assuming that SCION had been installed in that machine.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/86)


the4daspect:
Yes, it solves my issue. Thanks

PR #102, opened 2019-09-09
Update apps to new SCION version

### LINKED ISSUE #99: Deps.sh not supporting sudoless installations
Currently https://github.com/netsec-ethz/scion-apps/blob/master/deps.sh#L15 contains an explicit call to `sudo` which makes it impossible to execute the script in a sudoless environments, i.e. inside a docker.

This is a similar issue to https://github.com/scionproto/scion/issues/3065, i.e. this should be either explicitly stated in the docs if the environment has to be very specific (OS version, non-root user, whatever else) or the scripts should try to cover as generic use as possible.

Currently neither this nor scionproto docs say it's required to be non-root in a non-dockerized setup what is nowadays a more and more popular configuration.

chaehni:
## ready to merge *after* upgrade

TODO:

- [x] netcat
- [x]  ssh
- [x] test all apps with local / scionlab setup and all flags

Closes #99 

This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/102)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,8 +9,13 @@ all: $(TARGETS)
 deps:
 	./deps.sh
 
+vendor: deps
+
 clean:
 	@$(foreach d,$(SRCDIRS),cd $(ROOT_DIR)/$(d) && go clean;)
+
+mrproper: clean
+	rm -rf vendor/*/
 
 install: all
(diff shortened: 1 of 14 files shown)


juagargi:
Fixes #99

PR #103, opened 2019-09-10
Remove roughtime/

mkowalski:
As roughtime application is no longer maintained, does not run
as a service and is not planned to be used, it has been moved to
https://gitlab.inf.ethz.ch/OU-PERRIG/theses/msc_milan_pandurov and
is being removed from SCION-Apps which contains only maintained
codebase.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/103)


chaehni:
We should also remove roughtime from the Makefile.

PR #109, opened 2019-10-03
[webapp] WIP Latency Path Survey

mwfarb:
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/109)


mwfarb:
There are sill some bugs, but one question I have is the best way to show performance on the topology graph. Should we show the performance on the links only, or also show the inter AS performance on the graph?
Screenshot from 2019-10-03 08-45-04

mwfarb:
Another PR is more advanced than this one, so closing. See #124.

PR #111, opened 2019-10-25
Remove Wisc AP

mwfarb:
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/111)


mwfarb:
This should be followed up with the results of netsec-ethz/scion-tutorials/issues/118, when closed.

mkowalski:
Just FYI, we should be having bandwidth tester already running in all the APs (and many other infra nodes). Sorry for the confusion but the missing feature has been only implemented a few days ago :)

PR #117, opened 2019-11-28
Added support for user-defined path policies in SSH

radwasherif:
Allows SCION-SSH to make use of path policies, as defined [here](https://github.com/scionproto/scion/blob/master/doc/PathPolicy.md). The implementation is modular enough to be used for other SCION applications if needed. 

### Main additions: 

- ssh/conn_wrapper.go: wrapper around SCIONConn to allow for writing packets to paths that satisfy user-defined policy
- ssh/appconf.go : contains path policy, path selection mode, and optional storage of static path and static overlay in case of staic path selection mode
- ssh/client/main.go: new command-line arguments (as specified below), distinction between using `default.sock` for SCIONLab and `[ia].sock` for local topology.
- ssh/scionutils/scionutils.go: added `DialSCIONWithBindSVCWithConf`, which initializes the `ConnWrapper` with a `SCIONConn` and `AppConf` as parameters, and passes it to `QUIC`. 

```
./client [scion-options] -p <server port> <server address> [ssh options]

scion-options: 
	--policy-file, JSON file containing path policies
	--poilcy-name, name of path policy
	--selection [arbitrary | static | round-robin]
        --client-address, SCION address of ssh client host, for local topologies only.

```

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/117)


matzf:
@radwasherif @juagargi I made my comments _before_ Juan but forgot to "publish". Sorry if there is something that does not make chronological sense.

radwasherif:
___
*[ssh/conn_wrapper/conn_wrapper.go, line 1 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/117#-LurKn5xDy2VPaOfktTp:-Lv9rl5oBCTdQFWq4apd:b-jv060k) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/5e914222a6c2d3f99509080059d629af9db1c208/ssh/conn_wrapper/conn_wrapper.go#L1)):*
<details><summary><i>Previously, juagargi (Juan A. García Pardo Giménez de los Galanes) wrote…</i></summary><blockquote>

@matzf I agree with not organizing the library now. But I would place the functionality in an external library, even now that we don't have any other usage than ssh. The idea has been to organize all the changes to be independent from the ssh application, and to extend the usage of the current `snet` library functionality, while preserving the API.
</blockquote></details>

I could move it back to `ssh/scionutils` for now

radwasherif:
___
*[lib/scionutil/conn_wrapper.go, line 28 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/117#-Lv4gNx050LEEo5IWXwG:-Lv9tt5s8vOGUaEOC7Ad:bgjsvlq) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/b97f36bb34ec19b3649218ad70e8577ec27e4baa/lib/scionutil/conn_wrapper.go#L28)):*

ConnWrapper is a far too generic name, and so is AppConf. 
Maybe try to find a name that describes what this is for. OR move both to a separate package as suggested above, then this would be clear from context.

</blockquote></details>

As mentioned above, I'll move them back to `ssh/scionutils` for now. I don't like the names either. I'll think of something better.

radwasherif:
___
*[ssh/conn_wrapper/conn_wrapper.go, line 34 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/117#-LurQiTy2bTCIQnTch_t:-Lv9uHA89PHhHZikxSST:bwrenpa) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/5e914222a6c2d3f99509080059d629af9db1c208/ssh/conn_wrapper/conn_wrapper.go#L34)):*
<details><summary><i>Previously, juagargi (Juan A. García Pardo Giménez de los Galanes) wrote…</i></summary><blockquote>

@matzf I see your comment above about snet.SCIONConn not necessarily being public and agree.
</blockquote></details>

I rely on `snet.SCIONConn` being public and usable as an embedded type to avoid having to implement the entire interface and also make use of private methods etc.

radwasherif:
___
*[lib/scionutil/conn_wrapper.go, line 40 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/117#-Lv4qz2F4JKQ5VvCKeZu:-LvA-R9R7mrGFTUxtXqz:b-hd1j48) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/b97f36bb34ec19b3649218ad70e8577ec27e4baa/lib/scionutil/conn_wrapper.go#L40)):*
<details><summary><i>Previously, juagargi (Juan A. García Pardo Giménez de los Galanes) wrote…</i></summary><blockquote>

instead of having different behaviors depending on a switch, use normal OOP and just have different methods for different types (e.g. `staticConnWrapper`, `arbitraryConnWrapper` and so on).
</blockquote></details>

Alright, in this case it makes sense that `connWrapper` would be created by the `AppConf` I guess.

radwasherif:
___
*[lib/scionutil/conn_wrapper.go, line 40 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/117#-Lv4hwFZ8VbmVytpkP0q:-LvA-9FX-fYvNtu-63_2:b-wsynei) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/b97f36bb34ec19b3649218ad70e8577ec27e4baa/lib/scionutil/conn_wrapper.go#L40)):*

This is overriding only one of three `Write` methods that `snet.Conn` offers. The others will bypass your logic. Instead of offering the full `snet.Conn` interface, `connWrapper` could just implement `net.PacketConn` and you're good.
</blockquote></details>

My understanding is that applications running over QUIC call `WriteTo` because the remote address is never fixed anyway. You're right I'll probably mention this.

radwasherif:
___
*[lib/scionutil/conn_wrapper.go, line 40 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/117#-Lv4kyxC0aJzLz-LD1x_:-Lv9zpDXCX2HUP9Kk1yZ:b-gd21rx) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/b97f36bb34ec19b3649218ad70e8577ec27e4baa/lib/scionutil/conn_wrapper.go#L40)):*

Also, you seem to assume that every write is to the same address. That's probably be good enough for your usecase, but you should probably mention it.
</blockquote></details>

In light of my comment above, I'm not sure I understand what you mean here. A different address can be passed on every call, right?

juagargi:
___
*[ssh/scionutils/scionutils.go, line 6 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/117#-LwIpQoX0hzpC91H6J2J:-LwIpQoX0hzpC91H6J2K:b-utucb1) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/23eb262e1c452d1dcf6512b34174a4af9bae16ac/ssh/scionutils/scionutils.go#L6)):*
> ```Go
> 	"crypto/tls"
> 	"fmt"
> 	"github.com/netsec-ethz/netsec-scion/go/lib/common"
> ```

don't use `netsec-ethz/scion`. Every import of scion is `github.com/scionproto/scion/whatever`, as in the other lines. The `go.mod` file correctly  replaces all of them with our version.

juagargi:
___
*[ssh/scionutils/scionutils.go, line 32 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/117#-LwIpCxi73zEDcZD9iYb:-LwIpCxi73zEDcZD9iYc:b-g7mkz4) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/23eb262e1c452d1dcf6512b34174a4af9bae16ac/ssh/scionutils/scionutils.go#L32)):*
> ```Go
> 
> // DialSCION dials a SCION host and opens a new QUIC stream
> func DialSCION(localAddress string, remoteAddress string) (*quicconn.QuicConn, error) {
> ```

you have changed the signature of the function, but not all its callers. Build is broken!

PR #129, opened 2020-02-11
Update scion, remove snet.Addr

matzf:
Update to last version before monolith, replace snet.Addr with snet.UDPAddr.

* Extract rains name resolution to separate file (simplifies building without rains)
* Simplify hostfiles tests

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/129)


matzf:
Note: depends on netsec-ethz/rains#233

PR #139, opened 2020-04-07
Cleanup help texts and READMEs

matzf:
* remove sections talking about creating dummy openssl certificates
* cleanup netcat help text
* cleanup bwtestclient help text

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/139)


bwtester/bwtestclient/bwtestclient.go
```diff
@@ -62,26 +62,20 @@ func prepareAESKey() []byte {
...
 		"packet size.")
 	fmt.Println("\tSupported bandwidth unit prefixes are: none (e.g. 1500bps for 1.5kbps), k, M, G, T.")
 	fmt.Println("\tYou can also only set the target bandwidth, e.g. -cs 1Mbps")
-	fmt.Println("-sc specifies time duration, packet size, number of packets, target bandwidth of server->client " +
-		"test")
-	fmt.Println("\tYou can also only set the target bandwidth, e.g. -sc 1500kbps")
 	fmt.Println("\tWhen only the cs or sc flag is set, the other flag is set to the same value.")
-	fmt.Println("-i specifies if the client is used in interactive mode, " +
-		"when true the user is prompted for a path choice")
-	fmt.Println("Default test parameters are: ", DefaultBwtestParameters)
```
FR4NK-W, line 84: >fmt.Println("Default test parameters are: ", DefaultBwtestParameters)

I would keep the output of the default test parameters (even, or especially since they are arbitrary) in the help

ssh/README.md
```diff
@@ -2,60 +2,40 @@
...
-# Installation
-
-## Prerequisite
+### Usage
 
 SCION infrastructure has to be installed and running. Instructions can be found [here](https://netsec-ethz.github.io/scion-tutorials/)
 
-Additional development library for PAM is needed:
-```
-sudo apt-get install libpam0g-dev
```
FR4NK-W, line 13: This libpam is still a requirement for building this ssh app, I'd keep this info here as a reminder.

bwtester/bwtestclient/bwtestclient.go
matzf: The default value is part of the flag description printed with `flag.PrintDefaults()`. The full output is below. Do you agree that this does not need to be duplicated?

```
$ bwtestclient -h
Usage of bwtestclient:
  -cs string
        Client->Server test parameter (default "3,1000,30,80kbps")
  -i    Interactive path selection, prompt to choose path
  -pathAlgo string
        Path selection algorithm / metric ("shortest", "mtu")
  -s string
        Server SCION Address
  -sc string
        Server->Client test parameter (default "3,1000,30,80kbps")

Test parameters:
        -cs and -sc specifiy time duration (seconds), packet size (bytes), number of packets, and target bandwidth.
        The question mark character ? can be used as wildcard when setting the test parameters and its value is computed according to the other parameters. When more than one wilcard is used, all but the last one are set to the default values, e.g. ?,1000,?,5Mbps will run the test for the default duration and send as many packets as required to reach a bandwidth of 5 Mbps with the given packet size.
        Supported bandwidth unit prefixes are: none (e.g. 1500bps for 1.5kbps), k, M, G, T.
        You can also only set the target bandwidth, e.g. -cs 1Mbps
        You can also only set the target bandwidth, e.g. -sc 1500kbps
        When only the cs or sc flag is set, the other flag is set to the same value.
```

ssh/README.md
matzf: Done.

PR #142, opened 2020-05-03
Add SCION HTTP Proxy to shttp lib

martenwallewein:
Hi @all,

I've worked on setting up DVB-T video streaming over SCION and implemented a proxy component that can proxy requests from HTTP/1.1 to HTTP/3 over SCION and vice versa. We needed the proxy

1) To make the stream over SCION accessible by the vlc media player
2) To provide the DVB-T stream grabbed from the receiver via HTTP/3 oiver SCION

To better understant the setup, please consider the following example:

scion-video-setup

Considering the figure, the SCION HTTP Ingress Proxy was realised using the "toScion" direction and the Egress proxy using the "fromScion" direction as can be seen from the code.

It would be cool to have this implementation upstream, because we plan to integrate a demo into SCION apps that can play video streams using HTTP/3 over SCION considering a similar example as shown above, but play the stream in the browser instead of using vlc media player.

Please feel free to give me feedback on this PR :)

Best,
Marten

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/142)


martenwallewein:
Hi Matthias,

thanks for the feedback, I think is a more cleaner way to implement this and furthermore, the `setConfig` part can be moved outside of the lib itself. I'll give it a try!

martenwallewein:
I have two questions, maybe you can help me with them:
- is there an existing `isScionAddress` function? If not, could something like `parseUdpAddress` from https://github.com/scionproto/scion/blob/master/go/lib/snet/udpaddr.go help?
- I would add *tls.Config parameters to `shttp.ListenAndServe` and `shttp.NewSingleSCIONHostReverseProxy` so custom tls configuration/certs can be used, is it okay to add those parameters (especially for the listenAndServe method, I don't want to break exisiting implementations)

matzf:
> * is there an existing isScionAddress function? If not, could something like parseUdpAddress from https://github.com/scionproto/scion/blob/master/go/lib/snet/udpaddr.go help?

I don't think there is a public function like that, but you can use `snet.ParseUDPAddr` and check for errors, as you suggested.

> * I would add *tls.Config parameters to shttp.ListenAndServe and shttp.NewSingleSCIONHostReverseProxy so custom tls configuration/certs can be used, is it okay to add those parameters

Sure, that sounds good to me.

> * Should I squash my commits in the end?

You can if you like, but don't worry about it. We squash merge all pull requests anyway, so whatever history you have will be fine.

martenwallewein:
Hi Matthias, sorry for the late response. I added your feedback, compared to the initial implementation this one is now a lot smarter, thanks!

matzf:
Thanks @martenwallewein. If you could quickly update your branch, we'll get this merged :shipit:

martenwallewein:
Thanks, I updated the branch.

PR #156, opened 2020-07-28
Add example-shttp-fileserver

matzf:
Add example for simple fileserver that lists/serves all files under the current working directory.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/156)


cmeury:
I would have proposed to additions:
* Output requests and errors
* Provide the option to cache files in memory

matzf:
@cmeury #157

PR #162, opened 2020-08-12
mpsquic server: track reply path by client connection ID

matzf:
The server replies on the path on which the last packet was received.
Previously, this was indexed by the `(IA, IP, port)` tuple of the client, which kind of defeats the point of connection IDs; specifically, this did not work well with the connection setup race that the client does.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/162)


matzf:
This is broken and cannot be repaired; I should have read the docs more carefully. I had naively assumed that the connection IDs never change and can be tracked by simply looking at the visible packet headers during the initial handshake. However, the connection IDs can (and are) changed after the handshake with control frames in the encrypted traffic.

Issue #153, opened 2020-06-15
[webapp]: SCION WebView Idea


martenwallewein:
Hi all,
after working on the shttp proxy PR https://github.com/netsec-ethz/scion-apps/pull/142, we had the idea to bring our video streaming setups, based on this proxy, into scion-apps. To do so, I implemented a raw prototype to make HTTP/3 requests over SCION and display the results in scion-apps (as far as the browser can display them). This prototype looks like this:

![scion-webview](https://user-images.githubusercontent.com/32448709/84637863-81238d00-aef6-11ea-951e-c11b6c7ae4e2.png)

We plan to provide a DVB-T2 live stream over SCION that can be accessed via this feature. Furthermore, we hope that it could be an entrypoint for providing more resources via HTTP/3 over SCION. 

So before implementing this feature completely and creating a PR, I hope we can use this issue for feedback, discussion and opinions about it. 

Best,
Marten

matzf:
Hi Marten, thanks for this description and the prototype!

Having something visible in the browser would definitely be great.
Instead of building this as a feature of the webapp, I would perhaps first look at building a simple, stand-alone http proxy that can be directly used by the browser. Using a [Proxy auto-config](https://en.wikipedia.org/wiki/Proxy_auto-config) script, would make setup easy for users and allow to limit the proxy to requests that will actually use SCION.
A possible downside here would be that the browsers will reject SCION addresses as malformed URLs -- a simple workaround could be to just limit to using hostnames only, or devise a suitable scheme for the user to enter mangled addresses.

Note that this proxy would need to be a simple [forward proxy](https://en.wikipedia.org/wiki/Proxy_server#Types), not the _reverse_ proxy that you've added in #142.

What do you think?

martenwallewein:
Hi Matthias,

I like the idea of integrating SCION HTTP directly in the browser, not as part of scion-webapp, the proxy auto config approach looks promising. Maybe this can be combined with a small SCION browser extension, to allow scion addresses or maybe even a scion:// scheme, I know that this is possible, but I cannot oversee the effort to realize this. Nevertheless, I think realizing at this approach would be very beneficial, integrating it into webapp may be the "plan-b", we will see. I will do some research and testing about that and write down the results here, thanks for the help!

martenwallewein:
I see, nice one! I also thought to start with this PAC approach, but didn't manage to take time for this. However, this one looks like a good starting point for further progress.

PR #167, opened 2020-10-04
[WIP] feature/SCION HTTP endpoints and Webview

martenwallewein:
Hey all,

The idea of this PR is to provide easy access to resources shared via SCION HTTP. It's first mentioned in https://github.com/netsec-ethz/scion-apps/issues/153. Unfortunately, a solution that provides access to these resources directly in the browser is not really practicable, mainly because SCION addresses could not be handled properly. So my current approach would be to 

1) Allow the user to enter any SCION URL that can be accessed via SCION HTTP and display the results within the webapp
2) Collect some trending SCION HTTP resources that can be accessed directly (at the moment only our DVB-T2 live stream)

It looks like this (in a very basic version):

Bildschirmfoto vom 2020-10-04 14-19-51

Hopefully, we can collect more resources and finally make them easily accessible via webapp.

At the moment, this is only a raw prototype and WIP, a lot of things still have to be done before getting this upstream:
- Replace the current used proxy with some forward proxy that supports multiple proxied connections
- Remove the NextProtos flag or make them configurable
- Clean up code and add styles to UI
- Add information to default tabs (like required player for the live stream)


matzf:
C.f. https://github.com/netsec-ethz/scion-apps/issues/153

PR #179, opened 2020-12-23
[webapp] Updating to scion-apps v0.3.0

mmalesev:
It works with the local topology generated by scion/python/topology/generator.py, and with scionlab ASes. When running on a scionlab AS, scion log files do not exist any more, so some checks are skipped.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ build: scion-bat \
 	scion-sensorfetcher scion-sensorserver \
 	scion-skip \
 	scion-ssh scion-sshd \
+	scion-webapp \
 	example-helloworld \
 	example-hellodrkey \
 	example-shttp-client example-shttp-server example-shttp-fileserver example-shttp-proxy
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ setup_lint:
 
 lint:
 	@type golangci-lint > /dev/null || ( echo "golangci-lint not found. Install it manually or by running 'make setup_lint'."; exit 1 )
-	golangci-lint run --build-tags=$(TAGS)
+	golangci-lint run --build-tags=$(TAGS) --timeout=2m0s -v
 
 install: all
   # Note: install everything but the examples
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/mattn/go-sqlite3 v1.9.1-0.20180719091609-b3511bfdd742
 	github.com/msteinert/pam v0.0.0-20190215180659-f29b9f28d6f9
 	github.com/netsec-ethz/rains v0.2.0
+	github.com/pelletier/go-toml v1.8.1-0.20200708110244-34de94e6a887
 	github.com/scionproto/scion v0.6.0
 	github.com/smartystreets/goconvey v1.6.4
 	golang.org/x/crypto v0.0.0-20200820211705-5c72a883971a
--- a/webapp/README.md
+++ b/webapp/README.md
@@ -98,17 +98,17 @@ A list of dependencies for `webapp` can be found at [dependencies.md](./dependen
 ```shell
 Usage of webapp:
   -a string
-    	Address of server host. (default "127.0.0.1")
+        Address of server host. (default "127.0.0.1")
   -p int
-    	Port of server host. (default 8000)
+        Port of server host. (default 8000)
   -r string
-    	Root path to read/browse from, CAUTION: read-access granted from -a and -p. (default "$GOPATH/src/github.com/netsec-ethz/scion-apps/webapp/web/data")
+        Root path to read/browse from, CAUTION: read-access granted from -a and -p. (default "$GOPATH/src/github.com/netsec-ethz/scion-apps/webapp /web/data")
   -sgen string
-    	Path to read SCION gen directory of infrastructure config (default "/etc/scion")
+        Path to read SCION gen directory of infrastructure config (default "/etc/scion")
   -sgenc string
-    	Path to read SCION gen-cache directory of infrastructure run-time config (default "/var/lib/scion")
+        Path to read SCION gen-cache directory of infrastructure run-time config (default "/var/lib/scion")
   -srvroot string
-    	Path to read/write web server files. (default "$GOPATH/src/github.com/netsec-ethz/scion-apps/webapp/web")
+        Path to read/write web server files. (default "$GOPATH/src/github.com/netsec-ethz/scion-apps/webapp/web")
 ```
 
 ## Related Links
--- a/webapp/README.md
+++ b/webapp/README.md
@@ -61,19 +61,18 @@ Install all [SCIONLab apps](https://github.com/netsec-ethz/scion-apps) and depen
 ```shell
 cd scion-apps
 ./deps.sh
-make install
+make build
 ```
 
 Development Run on Local Topology:
 You can alter the defaults on the command line, all of which are listed below:
 ```shell
-webapp \
+./bin/scion-webapp \
 -a 127.0.0.1 \
 -p 8081 \
 -r . \
 -srvroot $GOPATH/src/github.com/netsec-ethz/scion-apps/webapp/web \
 -sabin $GOPATH/bin \
--sroot $GOPATH/src/github.com/scionproto/scion \
 -sbin $GOPATH/src/github.com/scionproto/scion/bin \
 -sgen $GOPATH/src/github.com/scionproto/scion/gen \
 -sgenc $GOPATH/src/github.com/scionproto/scion/gen-cache \
@@ -81,20 +80,19 @@ webapp \
 ```
 or can you run `webapp` like this, which will use the defaults above:
 ```shell
-webapp
+./bin/scion-webapp
 ```
 
 Development Run on SCIONLab Topology:
 ```shell
-webapp \
+./bin/scion-webapp \
 -a 0.0.0.0 \
 -p 8080 \
 -r ~/go/src/github.com/netsec-ethz/scion-apps/webapp/web/data \
 -srvroot ~/go/src/github.com/netsec-ethz/scion-apps/webapp/web \
 -sabin $GOPATH/bin \
--sroot /etc/scion \
 -sbin /usr/bin \
--sgen /etc/scion/gen \
+-sgen /etc/scion \
 -sgenc /var/lib/scion \
 -slogs /var/log/scion
 ```
@@ -121,8 +119,6 @@ Usage of webapp:
         Path to read SCION gen-cache directory of infrastructure run-time config (default "/home/ubuntu/go/src/github.com/scionproto/scion/gen-cache")
   -slogs string
         Path to read SCION logs directory of infrastructure logging (default "/home/ubuntu/go/src/github.com/scionproto/scion/logs")
-  -sroot string
-        Path to read SCION root directory of infrastructure (default "/home/ubuntu/go/src/github.com/scionproto/scion")
   -srvroot string
         Path to read/write web server files. (default "/home/ubuntu/go/src/github.com/netsec-ethz/scion-apps/webapp/web")
 ```
--- a/webapp/README.md
+++ b/webapp/README.md
@@ -64,37 +64,31 @@ cd scion-apps
 make build
 ```
 
-Development Run on Local Topology:
+Development Run on ScionLab Topology:
 You can alter the defaults on the command line, all of which are listed below:
 ```shell
 ./bin/scion-webapp \
 -a 127.0.0.1 \
--p 8081 \
--r . \
+-p 8000 \
+-r $GOPATH/src/github.com/netsec-ethz/scion-apps/webapp/web/data \
 -srvroot $GOPATH/src/github.com/netsec-ethz/scion-apps/webapp/web \
--sabin $GOPATH/bin \
--sbin $GOPATH/src/github.com/scionproto/scion/bin \
--sgen $GOPATH/src/github.com/scionproto/scion/gen \
--sgenc $GOPATH/src/github.com/scionproto/scion/gen-cache \
--slogs $GOPATH/src/github.com/scionproto/scion/logs
+-sgen /etc/scion \
+-sgenc /var/lib/scion 
 ```
 or can you run `webapp` like this, which will use the defaults above:
 ```shell
 ./bin/scion-webapp
 ```
 
-Development Run on SCIONLab Topology:
+Development Run on Local Topology:
 ```shell
 ./bin/scion-webapp \
--a 0.0.0.0 \
+-a 127.0.0.1 \
 -p 8080 \
--r ~/go/src/github.com/netsec-ethz/scion-apps/webapp/web/data \
--srvroot ~/go/src/github.com/netsec-ethz/scion-apps/webapp/web \
--sabin $GOPATH/bin \
--sbin /usr/bin \
--sgen /etc/scion \
--sgenc /var/lib/scion \
--slogs /var/log/scion
+-r $GOPATH/src/github.com/netsec-ethz/scion-apps/webapp/web/data \
+-srvroot $GOPATH/src/github.com/netsec-ethz/scion-apps/webapp/web \
+-sgen $HOME/scion/gen \
+-sgenc  $HOME/scion/gen-cache
 ```
 
 ## Dependencies
@@ -104,23 +98,17 @@ A list of dependencies for `webapp` can be found at [dependencies.md](./dependen
 ```shell
 Usage of webapp:
   -a string
-        Address of server host. (default "127.0.0.1")
+    	Address of server host. (default "127.0.0.1")
   -p int
-        Port of server host. (default 8000)
+    	Port of server host. (default 8000)
   -r string
-        Root path to read/browse from, CAUTION: read-access granted from -a and -p. (default ".")
-  -sabin string
-        Path to execute the installed scionlab apps binaries (default "/home/ubuntu/go/bin")
-  -sbin string
-        Path to execute SCION bin directory of infrastructure tools (default "/home/ubuntu/go/src/github.com/scionproto/scion/bin")
+    	Root path to read/browse from, CAUTION: read-access granted from -a and -p. (default "$GOPATH/src/github.com/netsec-ethz/scion-apps/webapp/web/data")
   -sgen string
-        Path to read SCION gen directory of infrastructure config (default "/home/ubuntu/go/src/github.com/scionproto/scion/gen")
+    	Path to read SCION gen directory of infrastructure config (default "/etc/scion")
   -sgenc string
-        Path to read SCION gen-cache directory of infrastructure run-time config (default "/home/ubuntu/go/src/github.com/scionproto/scion/gen-cache")
-  -slogs string
-        Path to read SCION logs directory of infrastructure logging (default "/home/ubuntu/go/src/github.com/scionproto/scion/logs")
+    	Path to read SCION gen-cache directory of infrastructure run-time config (default "/var/lib/scion")
   -srvroot string
-        Path to read/write web server files. (default "/home/ubuntu/go/src/github.com/netsec-ethz/scion-apps/webapp/web")
+    	Path to read/write web server files. (default "$GOPATH/src/github.com/netsec-ethz/scion-apps/webapp/web")
 ```
 
 ## Related Links
--- a/webapp/dependencies.md
+++ b/webapp/dependencies.md
@@ -28,7 +28,7 @@ webapp -h
 - systemctl
 
 ## SCION Binaries Used/Checked
-- $sbin/scmp
+- $sbin/scion (for scion ping and scion traceroute)
 - $sabin/bwtestclient
 - $sabin/imagefetcher
 - $sabin/sensorfetcher
--- a/webapp/dependencies.md
+++ b/webapp/dependencies.md
@@ -27,22 +27,21 @@ webapp -h
 - cat
 - systemctl
 
-## SCION Binaries Used/Checked
-- $sbin/scion (for scion ping and scion traceroute)
-- $sabin/bwtestclient
-- $sabin/imagefetcher
-- $sabin/sensorfetcher
+## SCION Binaries Used/Checked (should be somewhere on $PATH)
+- scion (for scion ping and scion traceroute)
+- scion-bwtestclient
+- scion-imagefetcher
+- scion-sensorfetcher
 
 ## Scripts/Directories Used/Checked
-- $sroot/scion.sh (deprecated)
-- $slogs/bs[IA]-1.log
-- $sgen/ISD*/AS* (scanning for local IAs)
-- $sgen/ISD[ISD]/AS[AS]/endhost/topology.json
+- $SCION_ROOT/scion.sh (only for local topologies (ISD < 16))
+- $sgen/.*topology.json (checking all subdirectories of $sgen for topology.json)
+- $sgen/.*(sciond|sd).toml (for reading sciond address)
+- $sgen/*cs*.toml (for reading prometheus server address)
 - $sgenc/*.crt
 - $sgenc/*.trc
 - $sgenc/ps[IA]-1.path.db
 - /run/shm/dispatcher/default.sock
-- /run/shm/sciond/default.sock
 
 ## Static Webserver Required
 - $srvroot/favico.ico
@@ -59,7 +58,3 @@ webapp -h
 - $srvroot/data/
 - $srvroot/data/images/
 - $srvroot/logs/
-
-
-
-
--- a/webapp/development.md
+++ b/webapp/development.md
@@ -111,10 +111,6 @@ gow run . \
 -p 8080 \
 -r ./web/data \
 -srvroot ./web \
--sabin /usr/bin/scion \
--sroot /etc/scion \
--sbin /usr/bin \
 -sgen  /etc/scion/gen \
 -sgenc /var/lib/scion \
--slogs /var/log/scion
 ```
--- a/webapp/development.md
+++ b/webapp/development.md
@@ -74,6 +74,7 @@ Add alternate test forwarding port line in `Vagrantfile`:
 Update Go Paths:
 ```shell
 echo 'export GOPATH="$HOME/go"' >> ~/.profile
+echo 'export GOBIN="$GOPATH/bin"' >> ~/.profile
 echo 'export PATH="$HOME/.local/bin:$GOPATH/bin:/usr/local/go/bin:$PATH"' >> ~/.profile
 source ~/.profile
 mkdir -p "$GOPATH"
@@ -91,24 +92,18 @@ Build and install `scion-apps`:
 ```shell
 sudo apt install make gcc libpam0g-dev
 cd ~
-git clone -b scionlab https://github.com/netsec-ethz/scion-apps
+git clone -b master https://github.com/netsec-ethz/scion-apps
 cd scion-apps
+make setup_lint
 make install
 ```
 
-Download scionlab's fork of scion and build and install `sig`:
-```shell
-cd ~
-git clone -b scionlab https://github.com/netsec-ethz/scion
-go build -o $GOPATH/bin/sig ~/scion/go/sig/main.go
-```
-
-Install Go Watcher:
+(Optional) Install Go Watcher:
 ```shell
 go get -u github.com/mitranim/gow 
 ```
 
-Development Run:
+Development Run (check dependencies.md for the meaning of the flags):
 ```shell
 cd ~/scion-apps/webapp
 gow run . \
--- a/webapp/lib/config.go
+++ b/webapp/lib/config.go
@@ -24,7 +24,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -44,6 +43,8 @@ var cfgFileCliUser = "config/clients_user.json"
 var cfgFileSerUser = "config/servers_user.json"
 var cfgFileCliDef = "config/clients_default.json"
 var cfgFileSerDef = "config/servers_default.json"
+
+var topologyFile = "topology.json"
 
 // command argument constants
 var CMD_ADR = "a"
@@ -67,6 +68,10 @@ var DEF_SCIONDIR = path.Join(GOPATH, "src/github.com/scionproto/scion")
 type UserSetting struct {
 	MyIA      string `json:"myIa"`
 	SDAddress string `json:"sdAddress"`
+}
+
+type Topology struct {
+	IA string `json:"isd_as"`
 }
 
 type CmdOptions struct {
@@ -217,20 +222,13 @@ func ReadUserSetting(options *CmdOptions) UserSetting {
 // ScanLocalIAs will load list of locally available IAs
 func ScanLocalIAs(options *CmdOptions) []string {
 	var localIAs []string
-	var reIaFilePathCap = `\/ISD([0-9]+)\/AS(\w+)`
-	re := regexp.MustCompile(reIaFilePathCap)
 	var searchPath = options.ScionGen
 	filepath.Walk(searchPath, func(path string, f os.FileInfo, _ error) error {
-		if f != nil && f.IsDir() {
-			capture := re.FindStringSubmatch(path)
-			if len(capture) > 0 {
-				ia := capture[1] + "-" + capture[2]
-				ia = strings.Replace(ia, "_", ":", -1) // convert once
-				if !StringInSlice(localIAs, ia) {
-					log.Debug("Local IA Found:", "ia", ia)
-					localIAs = append(localIAs, ia)
-				}
-			}
+		if f != nil && f.Name() == topologyFile {
+			var t Topology
+			raw, _ := ioutil.ReadFile(path)
+			json.Unmarshal([]byte(raw), &t)
+			localIAs = append(localIAs, t.IA)
 		}
 		return nil
 	})
--- a/webapp/lib/config.go
+++ b/webapp/lib/config.go
@@ -106,11 +106,12 @@ func isFlagUsed(name string) bool {
 }
 
 func defaultStaticRoot() string {
-	return "webapp/web"
+	defaultWebappRoot := path.Join(os.Getenv("GOPATH"), "/src/github.com/netsec-ethz/scion-apps/webapp/")
+	return path.Join(defaultWebappRoot, "web")
 }
 
-func defaultBrowseRoot() string {
-	return "data"
+func defaultBrowseRoot(staticRoot string) string {
+	return path.Join(staticRoot, "data")
 }
 
 func defaultScionGen() string {
@@ -126,7 +127,7 @@ func ParseFlags() CmdOptions {
 	port := flag.Int(CMD_PRT, listenPortDef, "Port of server host.")
 	staticRoot := flag.String(CMD_WEB, defaultStaticRoot(),
 		"Path to read/write web server files.")
-	browseRoot := flag.String(CMD_BRT, defaultBrowseRoot(),
+	browseRoot := flag.String(CMD_BRT, defaultBrowseRoot(*staticRoot),
 		"Root path to read/browse from, CAUTION: read-access granted from -a and -p.")
 	scionGen := flag.String(CMD_SCG, defaultScionGen(),
 		"Path to read SCION gen directory of infrastructure config")
@@ -138,7 +139,7 @@ func ParseFlags() CmdOptions {
 		*staticRoot = defaultStaticRoot()
 	}
 	if !isFlagUsed(CMD_BRT) {
-		*browseRoot = defaultBrowseRoot()
+		*browseRoot = defaultBrowseRoot(*staticRoot)
 	}
 	if !isFlagUsed(CMD_SCG) {
 		*scionGen = defaultScionGen()
@@ -189,10 +190,16 @@ func getMetricsServer(searchPath string) string {
 // getSDFromSDTomlFile returns sciond address from sd.toml on the given path
 func getSDFromSDTomlFile(path string) string {
 	var sd string
-	if config, err := toml.LoadFile(path); err == nil {
-		sd = config.Get("sd.address").(string)
-	} else {
-		log.Error("toml file %s could not be read", path)
+	switch config, err := toml.LoadFile(path); err {
+	case nil:
+		sdAddr := config.Get("sd.address")
+		if sdAddr != nil {
+			sd = sdAddr.(string)
+			break
+		}
+		fallthrough
+	default:
+		log.Info(fmt.Sprintf("sciond address could not be read from toml file %s", path))
 		var ok bool
 		sd, ok = os.LookupEnv("SCION_DAEMON_ADDRESS")
 		if !ok {
--- a/webapp/lib/config.go
+++ b/webapp/lib/config.go
@@ -28,9 +28,10 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/BurntSushi/toml"
 	log "github.com/inconshreveable/log15"
 	. "github.com/netsec-ethz/scion-apps/webapp/util"
-	"github.com/scionproto/scion/go/lib/addr"
+	"github.com/scionproto/scion/go/lib/sciond"
 )
 
 // default params for localhost testing
@@ -44,8 +45,10 @@ var cfgFileCliUser = "config/clients_user.json"
 var cfgFileSerUser = "config/servers_user.json"
 var cfgFileCliDef = "config/clients_default.json"
 var cfgFileSerDef = "config/servers_default.json"
+var cfgFileSerIA = "config/servers_iaToSd.json"
 
 var topologyFile = "topology.json"
+var sdFile = "sd.toml"
 
 // command argument constants
 var CMD_ADR = "a"
@@ -71,9 +74,21 @@ type UserSetting struct {
 	SDAddress string `json:"sdAddress"`
 }
 
-type Topology struct {
+// topology holds the IA from topology.json
+type topology struct {
 	IA string `json:"isd_as"`
 }
+
+// sdTomlConfig holds the information from sd.toml
+type sdTomlConfig struct {
+	SD sdInfo `toml:"sd"`
+}
+
+// sdInfo holds the Sciond infomation from the sd field in sd.toml
+type sdInfo struct {
+	Address string `toml:"address"`
+}
+
 
 type CmdOptions struct {
 	Addr          string
@@ -189,13 +204,13 @@ func ParseFlags() CmdOptions {
 // WriteUserSetting writes the settings to disk.
 func WriteUserSetting(options *CmdOptions, settings *UserSetting) {
 	cliUserFp := path.Join(options.StaticRoot, cfgFileCliUser)
+	settingsJSON, _ := json.Marshal(settings)
 
 	// writing myIA means we have to retrieve sciond's tcp address too
 	// since sciond's address may be autognerated
-	config, err := LoadSciondConfig(options, settings.MyIA)
+	sd, err := LoadSciondConfig(options, settings.MyIA)
 	CheckError(err)
-	settings.SDAddress = config.SD.Address
-	settingsJSON, _ := json.Marshal(settings)
+	settings.SDAddress = sd
 
 	log.Info("Updating...", "UserSetting", string(settingsJSON))
 	err = ioutil.WriteFile(cliUserFp, settingsJSON, 0644)
@@ -215,21 +230,64 @@ func ReadUserSetting(options *CmdOptions) UserSetting {
 	return settings
 }
 
-// ScanLocalIAs will load list of locally available IAs
-func ScanLocalIAs(options *CmdOptions) []string {
-	var localIAs []string
+// ScanLocalSetting will load list of locally available IAs and their corresponding Scionds
+func ScanLocalSetting(options *CmdOptions) map[string]string {
+	iaToSd := make(map[string]string)
 	var searchPath = options.ScionGen
 	filepath.Walk(searchPath, func(path string, f os.FileInfo, _ error) error {
 		if f != nil && f.Name() == topologyFile {
-			var t Topology
-			raw, _ := ioutil.ReadFile(path)
-			json.Unmarshal([]byte(raw), &t)
-			localIAs = append(localIAs, t.IA)
+			sdPath := path[:len(path)-len(topologyFile)] + sdFile
+			ia := getIAFromTopologyFile(path)
+			sd := getSDFromSDTomlFile(sdPath)
+			iaToSd[ia] = sd
 		}
 		return nil
 	})
-	sort.Strings(localIAs)
-	return localIAs
+	return iaToSd
+}
+
+// WriteLocalSettings writes the ia to sciond mapping to disk
+func WriteLocalSettings(options *CmdOptions, iaToSD map[string]string) {
+	fp := path.Join(options.StaticRoot, cfgFileSerIA)
+	settingsJSON, _ := json.Marshal(iaToSD)
+	log.Info("Updating...", "IAs and Scionds", string(settingsJSON))
+	err := ioutil.WriteFile(fp, settingsJSON, 0644)
+	CheckError(err)
+}
+
+// ReadLocalSettings reads the ia to sciond mappings from disk
+func ReadLocalSettings(options *CmdOptions) map[string]string {
+	iaToSD := make(map[string]string)
+	fp := path.Join(options.StaticRoot, cfgFileSerIA)
+
+	raw, err := ioutil.ReadFile(fp)
+	CheckError(err)
+	log.Debug("ReadLocalSettings from saved", "settings", string(raw))
+	json.Unmarshal([]byte(raw), &iaToSD)
+
+	return iaToSD
+}
+
+// getSDFromSDTomlFile returns sciond address from sd.toml on the given path
+func getSDFromSDTomlFile(path string) string {
+	var config sdTomlConfig
+	if _, err := toml.DecodeFile(path, &config); err == nil {
+		return config.SD.Address
+	}
+	// if sd.toml is not present, read from the environment variable SCION_DAEMON_ADDRESS
+	sd, ok := os.LookupEnv("SCION_DAEMON_ADDRESS")
+	if !ok {
+		sd = sciond.DefaultAPIAddress
+	}
+	return sd
+}
+
+// getIAFromTopologyFile returns IA from topology.json on the given path
+func getIAFromTopologyFile(path string) string {
+	var t topology
+	raw, _ := ioutil.ReadFile(path)
+	json.Unmarshal([]byte(raw), &t)
+	return t.IA
 }
 
 // StringInSlice can check a slice for a unique string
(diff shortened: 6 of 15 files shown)


matzf (changes requested):
> Reviewed 12 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)* status: all files reviewed, 8 unresolved discussions (waiting on @mmalesev)

___
*[webapp/lib/config.go, line 83 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHbBMG4LS4ggF07YSR:-MQHbBMG4LS4ggF07YSS:bf196h9) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L83)):*
> ```Go
> 	StaticRoot    string
> 	BrowseRoot    string
> 	AppsRoot      string
> ```

I believe `AppsRoot` is no longer necessary and it's use can be replaced with `ScionBin`.
Ditto for the related `APPS_ROOT` and `SCION_BIN` in the healthcheck scripts.
___
*[webapp/lib/config.go, line 219 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHFdN13OXtN_TgjQYi:-MQHFdN13OXtN_TgjQYj:bi3umfn) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L219)):*
> ```Go
> 
> // ScanLocalIAs will load list of locally available IAs
> func ScanLocalIAs(options *CmdOptions) []string {
> ```

The approach to discover the sciond addresses seems unnecessarily complicated. First we find the available ASes by listing all `topology.json` files here in `ScanLocalIAs`. Later we somehow traverse the directory hierarchy again to find the corresponding `sd.toml` file using the `GetPathInGen` below and extract the address from it.

I think it would simplify things if we'd record the sciond address immediately when discovering the ASes. For this we can inspect the `sd.toml` file in the same directory as the `topology.json` file we're currently looking at. The return type of this function needs to be adapted to allow returning the IAs together with the sciond address.
___
*[webapp/lib/config.go, line 219 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHbvLm4gk8-MZGBZow:-MQHbvLm4gk8-MZGBZox:bke77yk) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L219)):*
> ```Go
> 
> // ScanLocalIAs will load list of locally available IAs
> func ScanLocalIAs(options *CmdOptions) []string {
> ```

Perhaps it would also be beneficial to directly record the "base path" for the configuration of the AS. See comments on the check bash scripts.
___
*[webapp/lib/config.go, line 219 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHbwRq9JN56VoAlyEQ:-MQHbwRrBNuvpPCvmlCP:bnni82h) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L219)):*
> ```Go
> 
> // ScanLocalIAs will load list of locally available IAs
> func ScanLocalIAs(options *CmdOptions) []string {
> ```

And while we're at it: the path for the sciond DB files can also be read from the sciond config file. This would make the `findDBFilename` function and the command line parameter for `ScionGenCache` unnecessary.
Not blocking, but it seems like this would not be too much work and would make things a bit simpler and cleaner.
___
*[webapp/lib/config.go, line 367 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHGGgECLqrfm7B0Hvy:-MQHGGgECLqrfm7B0Hvz:b-tdazrp) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L367)):*
> ```Go
> 
> // GetPathInGen returns the full path of the fileName in scionGen directory
> func GetPathInGen(scionGen, fileName, ia string) (string, error) {
> ```

This should become unnecessary, see comment on `ScanLocalIAs`. Remove this.
___
*[webapp/lib/sciond.go, line 80 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHGEUS2fyhRJCrtBrs:-MQHGEUS2fyhRJCrtBrt:b-tdazrp) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/sciond.go#L80)):*
> ```Go
> }
> 
> func LoadSciondConfig(options *CmdOptions, ia string) (sdTomlConfig, error) {
> ```

This should become unnecessary, see comment on `ScanLocalIAs`. Remove this.
___
*[webapp/web/tests/health/beaconstest.sh, line 15 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHGzHx7DQBKnyPYmUP:-MQHGzHy2cY15fuxAZmx:b-jdlcm6) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/web/tests/health/beaconstest.sh#L15)):*
> ```Shell
> 
> # format log file and beacons grep string (was .DEBUG)
> # TODO (mmalesev) It seems like log files are not available in scionlab, find a different way to do beacon test
> ```

A (much!) better approach is to inspect the metrics exposed by the services.

Just as an example, here's a snippet from the scionlab CI scripts which waits for _some_ beacons to arrive.
```
       while curl -s 127.0.0.1:30454/metrics | egrep -q 'control_beaconing_received_beacons_total{.*}.[0]+'; do sleep 1; done
```

The address for the metrics needs to be read from the CS toml config file (which is called either `cs-1.toml` or `cs-$ia-1.toml`).
As commented in the `netcheck` script below, it would be helpful if the infrastructure to invoke these scripts would pass the base AS directory to find this `cs-*.toml` file.

With this change, the log directories are no longer needed and the corresponding parameters can be removed.
___
*[webapp/web/tests/health/netcheck.sh, line 24 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHZ_Sp4mmIjFj47glh:-MQHZ_Sp4mmIjFj47gli:bn6bhof) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/web/tests/health/netcheck.sh#L24)):*
> ```Shell
>         topologyFile=$SCION_GEN/topology.json
>     fi
> fi
> ```

Instead of this copy-pasted logic to find the AS configuration directory over and over again, could we just pass it as parameter to the script instead? This parameter could probably replace passing the IA, as this appears to not be used for anything other than finding the correct directory.
See comments on `ScanLocalIAs`.

mmalesev (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)* status: 6 of 14 files reviewed, 8 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[webapp/lib/config.go, line 83 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHbBMG4LS4ggF07YSR:-MTgd9j97G6EBM-O2qWB:b2eetvh) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L83)):*

I believe `AppsRoot` is no longer necessary and it's use can be replaced with `ScionBin`.
Ditto for the related `APPS_ROOT` and `SCION_BIN` in the healthcheck scripts.

</blockquote></details>

I don't think this is the case. For example, in ScionLab ASes ScionBin would be `/usr/bin`, whereas AppsRoot could be any path where you build scion apps (for example `~/go/bin`).
___
*[webapp/lib/config.go, line 219 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHFdN13OXtN_TgjQYi:-MTgO9XZ8Q4ZbTs1GJas:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L219)):*

The approach to discover the sciond addresses seems unnecessarily complicated. First we find the available ASes by listing all `topology.json` files here in `ScanLocalIAs`. Later we somehow traverse the directory hierarchy again to find the corresponding `sd.toml` file using the `GetPathInGen` below and extract the address from it.

I think it would simplify things if we'd record the sciond address immediately when discovering the ASes. For this we can inspect the `sd.toml` file in the same directory as the `topology.json` file we're currently looking at. The return type of this function needs to be adapted to allow returning the IAs together with the sciond address.
</blockquote></details>

Done.
___
*[webapp/lib/config.go, line 219 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHbvLm4gk8-MZGBZow:-MTgOMWX5DVjSvzUrqIh:bfkr37s) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L219)):*

Perhaps it would also be beneficial to directly record the "base path" for the configuration of the AS. See comments on the check bash scripts.
</blockquote></details>

Done. I decided to record full paths to topology.json and sd.toml though since filenames don't seem to be the same everywhere (`sd.toml` in local topology, `sciond.toml` in ScionLab ASes).
___
*[webapp/lib/config.go, line 219 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHbwRq9JN56VoAlyEQ:-MTgeyFACaT1WZyjnGID:b5uglrj) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L219)):*

And while we're at it: the path for the sciond DB files can also be read from the sciond config file. This would make the `findDBFilename` function and the command line parameter for `ScionGenCache` unnecessary.
Not blocking, but it seems like this would not be too much work and would make things a bit simpler and cleaner.
</blockquote></details>

ScionLab ASes make this easy since they have the absolute path in `sd.toml`. However, local topologies (for example the one in the integration test) have just a relative path, and it's not straightforward to figure out the absolute path just from the toml file.
___
*[webapp/lib/config.go, line 367 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHGGgECLqrfm7B0Hvy:-MTgfbsl4rUMOb9hDB8d:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L367)):*

This should become unnecessary, see comment on `ScanLocalIAs`. Remove this.
</blockquote></details>

Done.
___
*[webapp/lib/sciond.go, line 80 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHGEUS2fyhRJCrtBrs:-MTgfeJg8JUFpj5ErxDQ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/sciond.go#L80)):*

This should become unnecessary, see comment on `ScanLocalIAs`. Remove this.
</blockquote></details>

Done.
___
*[webapp/web/tests/health/netcheck.sh, line 24 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHZ_Sp4mmIjFj47glh:-MTgfocI66cdJeqs7Os_:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/web/tests/health/netcheck.sh#L24)):*

Instead of this copy-pasted logic to find the AS configuration directory over and over again, could we just pass it as parameter to the script instead? This parameter could probably replace passing the IA, as this appears to not be used for anything other than finding the correct directory.
See comments on `ScanLocalIAs`.
</blockquote></details>

Done.

matzf (changes requested):
> Reviewed 6 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)* status: 12 of 14 files reviewed, 7 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[webapp/lib/config.go, line 83 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHbBMG4LS4ggF07YSR:-MTkpwhP-LSJjFX2ojst:b5xnpd7) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L83)):*
<details><summary><i>Previously, mmalesev wrote…</i></summary><blockquote>

I don't think this is the case. For example, in ScionLab ASes ScionBin would be `/usr/bin`, whereas AppsRoot could be any path where you build scion apps (for example `~/go/bin`).
</blockquote></details>

The division seems rather arbitrary. Maybe remove _both_ and just lookup the binaries in `PATH` -- if the binaries are in a non-standard location, just extend `PATH`, which is something everybody is familiar with.
___
*[webapp/lib/config.go, line 219 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHbvLm4gk8-MZGBZow:-MTksXsF1ImOD-KxtKbO:bmi8ppe) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L219)):*
<details><summary><i>Previously, mmalesev wrote…</i></summary><blockquote>

Done. I decided to record full paths to topology.json and sd.toml though since filenames don't seem to be the same everywhere (`sd.toml` in local topology, `sciond.toml` in ScionLab ASes).
</blockquote></details>

:100: 
___
*[webapp/lib/config.go, line 219 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHbwRq9JN56VoAlyEQ:-MTktEcH5V1V9XVl8KCb:bv3u67f) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L219)):*
<details><summary><i>Previously, mmalesev wrote…</i></summary><blockquote>

ScionLab ASes make this easy since they have the absolute path in `sd.toml`. However, local topologies (for example the one in the integration test) have just a relative path, and it's not straightforward to figure out the absolute path just from the toml file.
</blockquote></details>

Yeah ok, that's a bit nasty indeed. We could interpret the paths as relative to `SCION_ROOT` but that seems less than ideal, so might not be worth changing.

Interesting alternative (perhaps for a separate PR): the CS now has an actual http API to retrieve segments from. So we can replace this entire DB finding/loading thing with a an API request.
___
*[webapp/lib/config.go, line 75 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTlZ67b5lrMRrvWRpke:-MTlZ67b5lrMRrvWRpkf:b-rp4dsy) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f9fd7aef3e34c17664b2b691404852a95c55d148/webapp/lib/config.go#L75)):*
> ```Go
> type UserSetting struct {
> 	MyIA string `json:"myIa"`
> 	Info IAInfo `json:"info"`
> ```

This should not be part of the user settings. It also does not ever need to be persisted. See also comment on `Write/ReadLocalSettings`.
___
*[webapp/lib/config.go, line 98 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTkrDrtCApCDpiCdrzw:-MTkrDrtCApCDpiCdrzx:bfgaebn) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f9fd7aef3e34c17664b2b691404852a95c55d148/webapp/lib/config.go#L98)):*
> ```Go
> type sdInfo struct {
> 	Address string `toml:"address"`
> }
> ```

We can get rid of these helper struts.
- use `github.com/pelletier/go-toml/` instead of (unmaintained) `BurntSushi/toml`
- read values directly: `config := toml.LoadFile(...); config.Get("sd.address").(string)`. See usage example (https://github.com/pelletier/go-toml/#usage-example)
   
___
*[webapp/lib/config.go, line 269 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTlXo3W8RH-eSIqayS8:-MTlXo3W8RH-eSIqayS9:b-tgxxj8) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f9fd7aef3e34c17664b2b691404852a95c55d148/webapp/lib/config.go#L269)):*
> ```Go
> 
> // ReadLocalSettings reads the ia to sciond mappings from disk
> func ReadLocalSettings(options *CmdOptions) map[string]IAInfo {
> ```

Wait, why? This does not need to be persisted at all. Maybe I'm misunderstanding, but it looks this is only here to pass this information from the `initLocalIAOptions` to the individual page handlers. 
Please remove these functions and instead pass this information to the individual handlers (e.g. pass `iaToInfo` to the handler function in `getPathInfoHandler`).
___
*[webapp/lib/sciond.go, line 71 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTlZpeH3aC8rc3Ng5HR:-MTlZpeIEbU6aqFkJCMo:bvu2vu8) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f9fd7aef3e34c17664b2b691404852a95c55d148/webapp/lib/sciond.go#L71)):*
> ```Go
> }
> 
> func GetSciondAddress(options *CmdOptions, ia string) (string, error) {
> ```

This function should be removed, see comment on `Write/ReadLocalSettings`.

mmalesev (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)* status: 12 of 14 files reviewed, 7 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[webapp/lib/config.go, line 219 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHbwRq9JN56VoAlyEQ:-MTllzlc6ALN_btL3JB3:bxfnu9j) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L219)):*

Yeah ok, that's a bit nasty indeed. We could interpret the paths as relative to `SCION_ROOT` but that seems less than ideal, so might not be worth changing.

Interesting alternative (perhaps for a separate PR): the CS now has an actual http API to retrieve segments from. So we can replace this entire DB finding/loading thing with a an API request.
</blockquote></details>

Also, it might be useful to use absolute paths in `sd.toml` for local topologies. It should be as simple as changing `db_dir` in [GoGenerator](https://github.com/scionproto/scion/blob/master/python/topology/go.py). Not sure where to put this `gen-cache` directory though.

mmalesev (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)* status: 6 of 15 files reviewed, 7 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[webapp/lib/config.go, line 83 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHbBMG4LS4ggF07YSR:-MTmpZFsAXNsOU3XEbIm:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L83)):*

The division seems rather arbitrary. Maybe remove _both_ and just lookup the binaries in `PATH` -- if the binaries are in a non-standard location, just extend `PATH`, which is something everybody is familiar with.
</blockquote></details>

Done.
___
*[webapp/lib/config.go, line 75 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTlZ67b5lrMRrvWRpke:-MTmc2c7D9mvEJV61Maz:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f9fd7aef3e34c17664b2b691404852a95c55d148/webapp/lib/config.go#L75)):*

This should not be part of the user settings. It also does not ever need to be persisted. See also comment on `Write/ReadLocalSettings`.
</blockquote></details>

Done.
___
*[webapp/lib/config.go, line 98 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTkrDrtCApCDpiCdrzw:-MTmc49X9Tm90JV7-9KZ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f9fd7aef3e34c17664b2b691404852a95c55d148/webapp/lib/config.go#L98)):*

We can get rid of these helper struts.
- use `github.com/pelletier/go-toml/` instead of (unmaintained) `BurntSushi/toml`
- read values directly: `config := toml.LoadFile(...); config.Get("sd.address").(string)`. See usage example (https://github.com/pelletier/go-toml/#usage-example)
   
</blockquote></details>

Done.
___
*[webapp/lib/config.go, line 269 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTlXo3W8RH-eSIqayS8:-MTmcCwICqfVgq1aNft4:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f9fd7aef3e34c17664b2b691404852a95c55d148/webapp/lib/config.go#L269)):*

Wait, why? This does not need to be persisted at all. Maybe I'm misunderstanding, but it looks this is only here to pass this information from the `initLocalIAOptions` to the individual page handlers. 
Please remove these functions and instead pass this information to the individual handlers (e.g. pass `iaToInfo` to the handler function in `getPathInfoHandler`).
</blockquote></details>

Done.
___
*[webapp/lib/sciond.go, line 71 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTlZpeH3aC8rc3Ng5HR:-MTmcVek6bumSrJFYA87:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f9fd7aef3e34c17664b2b691404852a95c55d148/webapp/lib/sciond.go#L71)):*

This function should be removed, see comment on `Write/ReadLocalSettings`.
</blockquote></details>

Done.
___
*[webapp/web/tests/health/beaconstest.sh, line 15 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHGzHx7DQBKnyPYmUP:-MTn5KAd47h7tYv8UJhO:bx54xh0) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/web/tests/health/beaconstest.sh#L15)):*

A (much!) better approach is to inspect the metrics exposed by the services.

Just as an example, here's a snippet from the scionlab CI scripts which waits for _some_ beacons to arrive.
```
       while curl -s 127.0.0.1:30454/metrics | egrep -q 'control_beaconing_received_beacons_total{.*}.[0]+'; do sleep 1; done
```

The address for the metrics needs to be read from the CS toml config file (which is called either `cs-1.toml` or `cs-$ia-1.toml`).
As commented in the `netcheck` script below, it would be helpful if the infrastructure to invoke these scripts would pass the base AS directory to find this `cs-*.toml` file.

With this change, the log directories are no longer needed and the corresponding parameters can be removed.
</blockquote></details>

Done. Local topology is a bit strange again (it's cs1-*.toml). Also, I'm not sure what would be a good test in this case so I just have a very simple one, basically just to check that the metrics server is running.

matzf (changes requested):
> Reviewed 1 of 8 files at r2, 8 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)* status: all files reviewed, 6 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[webapp/webapp.go, line 45 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTtinpj8UpvoeZ53agl:-MTtinpj8UpvoeZ53agm:b-lqrwzu) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/78c6ee8a85328085f5e8487a61c53523eb1fc3cb/webapp/webapp.go#L45)):*
> ```Go
> 
> // myIA is the currently selected IA
> var myIA string
> ```

Not for now, but wouldn't it be easier if we'd just remove this from the server state and explicitly pass it as parameter to all relevant  requests? It looks like this is already done for _some_ requests.
Maybe add a "FIXME" comment if you agree.
___
*[webapp/webapp.go, line 563 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTth5Cu6hg1diqjVXUe:-MTth5Cu6hg1diqjVXUf:b-dsnhqc) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/78c6ee8a85328085f5e8487a61c53523eb1fc3cb/webapp/webapp.go#L563)):*
> ```Go
> 			break
> 		}
> 	}
> ```

This is not wrong, but it's unnecessary. Just use the name of the binary (without any path) and `exec` will find it.
Just btw, there is a stdlib function `exec.LookPath`.
___
*[webapp/lib/config.go, line 83 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHbBMG4LS4ggF07YSR:-MTta0nb3-9r1WFgHi3K:bmi8ppe) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L83)):*
<details><summary><i>Previously, mmalesev wrote…</i></summary><blockquote>

Done.
</blockquote></details>

:100: 
___
*[webapp/lib/config.go, line 219 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHbwRq9JN56VoAlyEQ:-MTtaGpO9VsgpHNzAGt8:bz71n0g) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/lib/config.go#L219)):*
<details><summary><i>Previously, mmalesev wrote…</i></summary><blockquote>

Also, it might be useful to use absolute paths in `sd.toml` for local topologies. It should be as simple as changing `db_dir` in [GoGenerator](https://github.com/scionproto/scion/blob/master/python/topology/go.py). Not sure where to put this `gen-cache` directory though.
</blockquote></details>

This might not be welcome upstream, as it's convenient for development if this stuff refers to the local scion checkout (allows to move config between checkouts etc). The configuration is not *intended* to be inspected by an outside application... Let's just leave it like it is now, the problem will be gone once we switch to using the HTTP API instead of reading the DB files.
___
*[webapp/lib/config.go, line 75 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTlZ67b5lrMRrvWRpke:-MTtlj1u-la8UUno_pSw:bgtwm7v) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f9fd7aef3e34c17664b2b691404852a95c55d148/webapp/lib/config.go#L75)):*
<details><summary><i>Previously, mmalesev wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Just for the record: above, I did not mean to remove persisting the `MyIA` user settings. It does appear like this was a rather unnecessary complication, so I'm happy that you did anyway. :100: 
___
*[webapp/lib/config.go, line 269 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTlXo3W8RH-eSIqayS8:-MTtfQRJCy_lbM8yfO-7:b41ibcm) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f9fd7aef3e34c17664b2b691404852a95c55d148/webapp/lib/config.go#L269)):*
<details><summary><i>Previously, mmalesev wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Thanks! :100:
___
*[webapp/lib/config.go, line 67 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTtgVli9afV9-pzY7NQ:-MTtgVli9afV9-pzY7NR:b-8k39ou) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/78c6ee8a85328085f5e8487a61c53523eb1fc3cb/webapp/lib/config.go#L67)):*
> ```Go
> 
> // IAConfig holds data about ia
> type IAConfig struct {
> ```

Nit: I think `ASConfig` would be more correct.
___
*[webapp/lib/config.go, line 153 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTtmfh98RxGsMlJEO9k:-MTtmfhA3ylk0qQ18GCp:b-89jmvc) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/78c6ee8a85328085f5e8487a61c53523eb1fc3cb/webapp/lib/config.go#L153)):*
> ```Go
> // ScanLocalSetting will load list of locally available IAs and their corresponding Scionds
> func ScanLocalSetting(options *CmdOptions) map[string]IAConfig {
> 	iaToCfg := make(map[string]IAConfig)
> ```

Nits:
- add a type definition for this, `type ASConfigs = map[str]ASConfig`
- rename the var here and the global `asConfigs`
___
*[webapp/lib/config.go, line 176 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTtbcyb6TLgygGgLFfg:-MTtbcyb6TLgygGgLFfh:bshxenq) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/78c6ee8a85328085f5e8487a61c53523eb1fc3cb/webapp/lib/config.go#L176)):*
> ```Go
> }
> 
> // getMetricsServer returns metrics server address from sc-*.toml on the given path
> ```

typo: `cs-*.toml`
___
*[webapp/lib/config.go, line 181 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTtbloB9ahAKpw_Tz2d:-MTtbloB9ahAKpw_Tz2e:ba9a5al) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/78c6ee8a85328085f5e8487a61c53523eb1fc3cb/webapp/lib/config.go#L181)):*
> ```Go
> 	filepath.Walk(searchPath, func(path string, f os.FileInfo, _ error) error {
> 		// searching for cs*.toml file (a hacky way to avoid importing a regex package)
> 		if f != nil && len(f.Name()) > len("cs-.toml") && f.Name()[:2] == "cs" &&
> ```

Curious, regex is in the standard library, so why do you want to avoid it?
Anyway, just use `filepath.Glob`:
```go
files, err := filepath.Glob(filepath.Join(searchPath,  "cs-*.toml"))
if err != nil || len(files) != 1 {
   return ""
}
...
```
___
*[webapp/web/tests/health/beaconstest.sh, line 15 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHGzHx7DQBKnyPYmUP:-MTtkTiN7Deh0Az9uDuy:bw5gimv) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/web/tests/health/beaconstest.sh#L15)):*
<details><summary><i>Previously, mmalesev wrote…</i></summary><blockquote>

Done. Local topology is a bit strange again (it's cs1-*.toml). Also, I'm not sure what would be a good test in this case so I just have a very simple one, basically just to check that the metrics server is running.
</blockquote></details>

That does not seem very useful, the `testSCIONrunning` script already seems to cover this.
The check I suggested above checks that beacons are arriving (which is what this check was previously supposed to do).

mmalesev (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)* status: all files reviewed, 5 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[webapp/webapp.go, line 45 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTtinpj8UpvoeZ53agl:-MTvQh5i9p2UZYL0vZtk:b54v6ey) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/78c6ee8a85328085f5e8487a61c53523eb1fc3cb/webapp/webapp.go#L45)):*

Not for now, but wouldn't it be easier if we'd just remove this from the server state and explicitly pass it as parameter to all relevant  requests? It looks like this is already done for _some_ requests.
Maybe add a "FIXME" comment if you agree.
</blockquote></details>

It doesn't seem that straightforward to me. For example, I don't see an easy way to pass it to the http handler functions.
___
*[webapp/webapp.go, line 563 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTth5Cu6hg1diqjVXUe:-MTvVS0DEzFVCAOHsBFL:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/78c6ee8a85328085f5e8487a61c53523eb1fc3cb/webapp/webapp.go#L563)):*

This is not wrong, but it's unnecessary. Just use the name of the binary (without any path) and `exec` will find it.
Just btw, there is a stdlib function `exec.LookPath`.
</blockquote></details>

Done.
___
*[webapp/lib/config.go, line 153 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTtmfh98RxGsMlJEO9k:-MTvXed40uvv0-tcROFg:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/78c6ee8a85328085f5e8487a61c53523eb1fc3cb/webapp/lib/config.go#L153)):*

Nits:
- add a type definition for this, `type ASConfigs = map[str]ASConfig`
- rename the var here and the global `asConfigs`
</blockquote></details>

Done.
___
*[webapp/lib/config.go, line 176 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTtbcyb6TLgygGgLFfg:-MTvZzsWBOMeoj35lo8E:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/78c6ee8a85328085f5e8487a61c53523eb1fc3cb/webapp/lib/config.go#L176)):*

typo: `cs-*.toml`
</blockquote></details>

Done.
___
*[webapp/lib/config.go, line 181 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTtbloB9ahAKpw_Tz2d:-MTvagw16XUONOw03fIq:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/78c6ee8a85328085f5e8487a61c53523eb1fc3cb/webapp/lib/config.go#L181)):*

Curious, regex is in the standard library, so why do you want to avoid it?
Anyway, just use `filepath.Glob`:
```go
files, err := filepath.Glob(filepath.Join(searchPath,  "cs-*.toml"))
if err != nil || len(files) != 1 {
   return ""
}
...
```
</blockquote></details>

Done.
___
*[webapp/web/tests/health/beaconstest.sh, line 15 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHGzHx7DQBKnyPYmUP:-MTvc5V14hBNqbAJVlYA:beldvcn) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/web/tests/health/beaconstest.sh#L15)):*

That does not seem very useful, the `testSCIONrunning` script already seems to cover this.
The check I suggested above checks that beacons are arriving (which is what this check was previously supposed to do). 
</blockquote></details>

I didn't use that one because `control_beaconing_received` does not appear in the metrics for local topologies, but I don't see any errors.

matzf (changes requested):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)* status: 10 of 15 files reviewed, 5 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[webapp/web/tests/health/beaconstest.sh, line 15 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHGzHx7DQBKnyPYmUP:-MUDDBV95wm2KoenoeQp:b-smm80y) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/web/tests/health/beaconstest.sh#L15)):*
<details><summary><i>Previously, mmalesev wrote…</i></summary><blockquote>

I didn't use that one because `control_beaconing_received` does not appear in the metrics for local topologies, but I don't see any errors.
</blockquote></details>

Yes, it does, output from default topology below. Note that it only shows up when it's non-zero, and so it may not ever show up if an AS does not receive beacons (e.g. core AS in tiny topo without core links). So, maybe this is a fair point, maybe `control_beaconing_registered_segments_total` (filter for `result="ok_.*`) might be a better indicator as I think it should become non-zero in any AS even in the tiny topology.

matzf (changes requested):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)* status: 10 of 15 files reviewed, 5 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[webapp/web/tests/health/beaconstest.sh, line 15 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHGzHx7DQBKnyPYmUP:-MUDFUAeArLZfWN3aMpM:b3i0eb5) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/web/tests/health/beaconstest.sh#L15)):*

Yes, it does, output from default topology below. Note that it only shows up when it's non-zero, and so it may not ever show up if an AS does not receive beacons (e.g. core AS in tiny topo without core links). So, maybe this is a fair point, maybe `control_beaconing_registered_segments_total` (filter for `result="ok_.*`) might be a better indicator as I think it should become non-zero in any AS even in the tiny topology.
</blockquote></details>

Forgot to include the example output above.
```bash
$ curl -sfS 127.0.0.59:30452/metrics | grep control_beaconing_received_beacons_total
# HELP control_beaconing_received_beacons_total Total number of beacons received.
# TYPE control_beaconing_received_beacons_total counter
control_beaconing_received_beacons_total{ingress_interface="494",neighbor_isd_as="1-ff00:0:111",result="ok_old"} 2
control_beaconing_received_beacons_total{ingress_interface="494",neighbor_isd_as="1-ff00:0:111",result="ok_updated"} 80
control_beaconing_received_beacons_total{ingress_interface="495",neighbor_isd_as="1-ff00:0:130",result="ok_updated"} 41
```

and for the `registered_segments`:
```bash
$ curl -sfS 127.0.0.59:30452/metrics | grep control_beaconing_registered_segments_total
# HELP control_beaconing_registered_segments_total Total number of segments registered.
# TYPE control_beaconing_registered_segments_total counter
control_beaconing_registered_segments_total{ingress_interface="494",result="err_network",seg_type="down",start_isd_as="1-ff00:0:120"} 9
control_beaconing_registered_segments_total{ingress_interface="494",result="err_network",seg_type="down",start_isd_as="1-ff00:0:130"} 10
control_beaconing_registered_segments_total{ingress_interface="494",result="ok_success",seg_type="down",start_isd_as="1-ff00:0:120"} 163
control_beaconing_registered_segments_total{ingress_interface="494",result="ok_success",seg_type="down",start_isd_as="1-ff00:0:130"} 162
control_beaconing_registered_segments_total{ingress_interface="494",result="ok_updated",seg_type="up",start_isd_as="1-ff00:0:120"} 151
control_beaconing_registered_segments_total{ingress_interface="494",result="ok_updated",seg_type="up",start_isd_as="1-ff00:0:130"} 146
control_beaconing_registered_segments_total{ingress_interface="495",result="err_network",seg_type="down",start_isd_as="1-ff00:0:130"} 10
control_beaconing_registered_segments_total{ingress_interface="495",result="ok_success",seg_type="down",start_isd_as="1-ff00:0:130"} 162
control_beaconing_registered_segments_total{ingress_interface="495",result="ok_updated",seg_type="up",start_isd_as="1-ff00:0:130"} 161
```

matzf (commented):
> Reviewed 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)* status: all files reviewed, 2 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[webapp/webapp.go, line 45 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MTtinpj8UpvoeZ53agl:-MUDIs741jlk8Riv7EZj:bb93mf1) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/78c6ee8a85328085f5e8487a61c53523eb1fc3cb/webapp/webapp.go#L45)):*
<details><summary><i>Previously, mmalesev wrote…</i></summary><blockquote>

It doesn't seem that straightforward to me. For example, I don't see an easy way to pass it to the http handler functions.
</blockquote></details>

I meant: pass the IA currently selected in the drop down menu as a parameter in each HTTP request from the browser to the webapp-server. Then the webapp server does not need to track `myIA`. The information is not _passed_ to the handler, the handler extracts it from the request data. Similar to how `ia_ser` and `ia_cli` are currently handled in `parseRequest2CmdItem`.
Anyway, not for now...

matzf (changes requested):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)* status: all files reviewed, 6 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[webapp/dependencies.md, line 31 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MUDL3fcAcDMagM_JyV9:-MUDL3fcAcDMagM_JyVA:b-mmxb8v) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/477859594e60cf1369734037afabea8c1a8148eb/webapp/dependencies.md#L31)):*
> ```Markdown
> 
> ## SCION Binaries Used/Checked
> - $sbin/scion (for scion ping and scion traceroute)
> ```

Nit: remove the `$sbin`/`$sabin` prefixes.
___
*[webapp/README.md, line 76 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MUDKkPe4lDYm_fk2pvZ:-MUDKkPe4lDYm_fk2pv_:b-ify0ac) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/477859594e60cf1369734037afabea8c1a8148eb/webapp/README.md#L76)):*
> ```Markdown
> -srvroot $GOPATH/src/github.com/netsec-ethz/scion-apps/webapp/web \
> -sabin $GOPATH/bin \
> -sbin $GOPATH/src/github.com/scionproto/scion/bin \
> ```

Some more cleanup: `-sabin`, `-sbin`,  and `-slogs` are gone. Perhaps mention that binaries are searched on `PATH`.
___
*[webapp/web/tests/health/netcheck.sh, line 17 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MUDLdNT40OjZr3u-GiA:-MUDLdNT40OjZr3u-GiB:b9fdddg) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/477859594e60cf1369734037afabea8c1a8148eb/webapp/web/tests/health/netcheck.sh#L17)):*
> ```Shell
> # find the topology file in gen
> isd=$(echo ${iaFile} | cut -d"-" -f1)
> as=$(echo ${iaFile} | cut -d"-" -f2)
> ```

Unused variables `iaFile`, `isd`, `as`.
___
*[webapp/web/tests/health/scmpcheck.sh, line 48 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MUDLtfR2UeNiq3JbM5O:-MUDLtfR2UeNiq3JbM5P:b-e9ks36) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/477859594e60cf1369734037afabea8c1a8148eb/webapp/web/tests/health/scmpcheck.sh#L48)):*
> ```Shell
>     if [ $isd -lt 16 ]; then
>         # local tests
>         cmd="$cmd --sciond $sdaddress"
> ```

Nit: append `--sciond`  unconditionally, then `isd`, `as`, `ia` and `iaFile` are unused and can be removed.
___
*[webapp/web/tests/health/testVPN.sh, line 35 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MUDMfIM6336hI5MfReF:-MUDMfIM6336hI5MfReG:b-1w3luo) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/477859594e60cf1369734037afabea8c1a8148eb/webapp/web/tests/health/testVPN.sh#L35)):*
> ```Shell
> interfaces=next(iter(brs.values()))['Interfaces']
> inter=next(iter(interfaces.values()))
> print(inter['BindOverlay'] if 'BindOverlay' in inter.keys() else inter['PublicOverlay']['Addr'])")
> ```

This looks like it won't work (old keys). 

This whole script does not make much sense anyway, though, and it's rather broken:
- the filename it's looking for (`/etc/openvpn/client.conf`) is not accurate, it should be `client-scionlab-*.conf` (and there can be multiple).
- simply looking for `tun0` is a _very_ coarse check. Perhaps check for the status of (each) openvpn@client-scionlab-....service instead?
- the python thing parsing the topology should report multiple addresses, and the script should check that for _each_ one there is a tun device with a matching IP. 
- the diagnostics messages are terrible ("Please destroy your VM", seriously? maybe advise to run `scionlab-config --force` instead)

Ok for me to make this an issue and address it in a separate PR.

mmalesev (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)* status: 8 of 15 files reviewed, 4 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[webapp/README.md, line 76 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MUDKkPe4lDYm_fk2pvZ:-MUEa1BQ-p0d2DRieQS0:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/477859594e60cf1369734037afabea8c1a8148eb/webapp/README.md#L76)):*

Some more cleanup: `-sabin`, `-sbin`,  and `-slogs` are gone. Perhaps mention that binaries are searched on `PATH`.
</blockquote></details>

Done.
___
*[webapp/web/tests/health/beaconstest.sh, line 15 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MQHGzHx7DQBKnyPYmUP:-MUEhJJOBwdhpF5rXRCu:b-1digaz) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f925b440e14f3b5dd881d44056a6197e20a7c190/webapp/web/tests/health/beaconstest.sh#L15)):*

Forgot to include the example output above.
```bash
$ curl -sfS 127.0.0.59:30452/metrics | grep control_beaconing_received_beacons_total
# HELP control_beaconing_received_beacons_total Total number of beacons received.
# TYPE control_beaconing_received_beacons_total counter
control_beaconing_received_beacons_total{ingress_interface="494",neighbor_isd_as="1-ff00:0:111",result="ok_old"} 2
control_beaconing_received_beacons_total{ingress_interface="494",neighbor_isd_as="1-ff00:0:111",result="ok_updated"} 80
control_beaconing_received_beacons_total{ingress_interface="495",neighbor_isd_as="1-ff00:0:130",result="ok_updated"} 41
```

and for the `registered_segments`:
```bash
$ curl -sfS 127.0.0.59:30452/metrics | grep control_beaconing_registered_segments_total
# HELP control_beaconing_registered_segments_total Total number of segments registered.
# TYPE control_beaconing_registered_segments_total counter
control_beaconing_registered_segments_total{ingress_interface="494",result="err_network",seg_type="down",start_isd_as="1-ff00:0:120"} 9
control_beaconing_registered_segments_total{ingress_interface="494",result="err_network",seg_type="down",start_isd_as="1-ff00:0:130"} 10
control_beaconing_registered_segments_total{ingress_interface="494",result="ok_success",seg_type="down",start_isd_as="1-ff00:0:120"} 163
control_beaconing_registered_segments_total{ingress_interface="494",result="ok_success",seg_type="down",start_isd_as="1-ff00:0:130"} 162
control_beaconing_registered_segments_total{ingress_interface="494",result="ok_updated",seg_type="up",start_isd_as="1-ff00:0:120"} 151
control_beaconing_registered_segments_total{ingress_interface="494",result="ok_updated",seg_type="up",start_isd_as="1-ff00:0:130"} 146
control_beaconing_registered_segments_total{ingress_interface="495",result="err_network",seg_type="down",start_isd_as="1-ff00:0:130"} 10
control_beaconing_registered_segments_total{ingress_interface="495",result="ok_success",seg_type="down",start_isd_as="1-ff00:0:130"} 162
control_beaconing_registered_segments_total{ingress_interface="495",result="ok_updated",seg_type="up",start_isd_as="1-ff00:0:130"} 161
```
</blockquote></details>

I've added `control_beaconing_registered_segments_total`, but this one also fails for coreAS 110 in tiny topo.
___
*[webapp/web/tests/health/netcheck.sh, line 17 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MUDLdNT40OjZr3u-GiA:-MUEaRQ86QsalhPa-wvC:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/477859594e60cf1369734037afabea8c1a8148eb/webapp/web/tests/health/netcheck.sh#L17)):*

Unused variables `iaFile`, `isd`, `as`.
</blockquote></details>

Done.
___
*[webapp/web/tests/health/testVPN.sh, line 35 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MUDMfIM6336hI5MfReF:-MUEcUqs6BRNBBza3lpW:b-wcmz9r) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/477859594e60cf1369734037afabea8c1a8148eb/webapp/web/tests/health/testVPN.sh#L35)):*

This looks like it won't work (old keys). 

This whole script does not make much sense anyway, though, and it's rather broken:
- the filename it's looking for (`/etc/openvpn/client.conf`) is not accurate, it should be `client-scionlab-*.conf` (and there can be multiple).
- simply looking for `tun0` is a _very_ coarse check. Perhaps check for the status of (each) openvpn@client-scionlab-....service instead?
- the python thing parsing the topology should report multiple addresses, and the script should check that for _each_ one there is a tun device with a matching IP. 
- the diagnostics messages are terrible ("Please destroy your VM", seriously? maybe advise to run `scionlab-config --force` instead)

Ok for me to make this an issue and address it in a separate PR.
</blockquote></details>

Fair, I will open a couple issues for stuff I don't have time for atm.

matzf (changes requested):
> Reviewed 7 of 7 files at r5.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)* status: all files reviewed, 1 unresolved discussion / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[webapp/lib/config.go, line 202 at r5](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MUFR0Ex07UOuWi4tz7D:-MUFR0EyEe6rVNEno5De:bviihsj) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/5853f77eb1b403629b2cc30c9ee7d0a403a3018c/webapp/lib/config.go#L202)):*
> ```Go
> 		fallthrough
> 	default:
> 		log.Info(fmt.Sprintf("sciond address could not be read from toml file %s", path))
> ```

This control flow seems quite puzzling now. Using `return` in your first approach seems to make this more natural:
```go
if config, err := toml.Load(path); err == nil {
   if sdAddr := config.Get("sd.address"); sdAddr != nil{
      return sdAddr.(string)
   }
}
// fallback:
log.Info(...)
if sd, ok := os.LookupEnv(...); ok {
   return sd
}
return sciond.DefaultAPIAddress
```

mmalesev (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)* status: 14 of 15 files reviewed, 1 unresolved discussion / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[webapp/lib/config.go, line 202 at r5](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-MUFR0Ex07UOuWi4tz7D:-MUFXPHACEiXvASsc8Sr:b-mygx3m) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/5853f77eb1b403629b2cc30c9ee7d0a403a3018c/webapp/lib/config.go#L202)):*

This control flow seems quite puzzling now. Using `return` in your first approach seems to make this more natural:
```go
if config, err := toml.Load(path); err == nil {
   if sdAddr := config.Get("sd.address"); sdAddr != nil{
      return sdAddr.(string)
   }
}
// fallback:
log.Info(...)
if sd, ok := os.LookupEnv(...); ok {
   return sd
}
return sciond.DefaultAPIAddress
```
</blockquote></details>

Done. I was just happy to finally use `fallthrough` somewhere :sweat_smile:

matzf (approved):
:lgtm:

> Reviewed 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179#-:-MUHzvbvCI41LCsO9kwB:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

matzf (approved):
> Reviewed 2 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)* status: :shipit: complete! all files reviewed, all discussions resolved

matzf (approved):
> Reviewed 1 of 1 files at r8.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/179)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #185, opened 2021-03-02
skip: bind address flag, serve skip.pac via http

matzf:
Add --bind address for local address. Uses kingpin to parse flags simply
because this library is already in use in this repo.

The skip.pac file is now interpreted as a template so that the local
address can be filled in and served on http (instead of previously just
installing it with a file://-URL).
For this, we use the new "go:embed" feature to embed the template file
into the binary (hence the update to go-1.16).

Configuring the skip.pac file with an http://-URL appears to make this
work in chrome :tada:.

Also:
- update to go-1.16 (for go:embed, see above), update/cleanup related do docs
- update golangci-lint version in Makefile
- go mod tidy

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/185)


marcfrei:
___
*[skip/main.go, line 68 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/185#-MUnOELBA7vWhZ2PAveE:-MUnOELBA7vWhZ2PAveF:bcbelpb) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f1aada8a9204dfa5fbc7de9026f1739ee78c1bc4/skip/main.go#L68)):*
> ```Go
> 	}
> 	mux := http.NewServeMux()
> 	mux.Handle("localhost/skip.pac", http.HandlerFunc(handleWPAD))
> ```

Is there a reason why we don't just register skip.pac at the root like
```
mux.Handle("/skip.pac", http.HandlerFunc(handleWPAD))
```

`localhost` doesn't work well in my setup, where I run skip in a local VM. But with skip.pac at the root it worked fine so far.

marcfrei (commented):
> Reviewed 10 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/185)* status: all files reviewed, 1 unresolved discussion / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

marcfrei:
:lgtm:

marcfrei (approved):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/185)* status: :shipit: complete! 1 of 1 LGTMs obtained

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/185)* status: 9 of 10 files reviewed, 1 unresolved discussion / 0 of 1 LGTMs obtained, and 1 stale / 0 of 1 approvals obtained

___
*[skip/main.go, line 68 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/185#-MUnOELBA7vWhZ2PAveE:-MUnQuyvExnJvSrZm2N4:b-s541zz) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f1aada8a9204dfa5fbc7de9026f1739ee78c1bc4/skip/main.go#L68)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

Is there a reason why we don't just register skip.pac at the root like
```
mux.Handle("/skip.pac", http.HandlerFunc(handleWPAD))
```

`localhost` doesn't work well in my setup, where I run skip in a local VM. But with skip.pac at the root it worked fine so far.
</blockquote></details>

Yes, otherwise this would make impossible to proxy any request matching `/skip.pac`. This would unlikely be an issue, but it feels just wrong to ignore it. It's a good point though, I forgot to revisit this when adding the --bind option. I've now added another route for the local bind IP address. Does that work for you?

marcfrei:
___
*[skip/main.go, line 68 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/185#-MUnOELBA7vWhZ2PAveE:-MUniYD59RBZblD5mufb:b1gd00f) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f1aada8a9204dfa5fbc7de9026f1739ee78c1bc4/skip/main.go#L68)):*

Yes, otherwise this would make impossible to proxy any request matching `/skip.pac`. This would unlikely be an issue, but it feels just wrong to ignore it. It's a good point though, I forgot to revisit this when adding the --bind option. I've now added another route for the local bind IP address. Does that work for you? 
</blockquote></details>

Works fine now. Thanks! I was lazy and used 0.0.0.0 as the local bind address and then it obviously still doesn't work. But binding to a specific address seems to be cleaner anyway and I think this is a good solution now.

marcfrei (approved):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/185)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #187, opened 2021-03-24
Policy-based path-aware SCION application library

matzf:
## Policy-based, path-aware SCION application library

Working title "pan" (for pan-ready Path-Aware Networking library (?)).

The goal is to have a common, high-level library that makes it easy to write correct, functional applications using SCION.

### Overview

The main entry points for applications are:

 - DialUDP / ListenUDP
 - DialQUIC / ListenQUIC

Both forms of the Dial call allow to specify a Policy and a Selector.

#### Policy
A path policy defines the allowed paths and/or a preference order of the paths. Policies are generally stateless and, in particular, they don't look for any short term information like measured latency or path "liveness".
Connections allow to change the path policy at any time.

Built in policies:
* [x] Interactive selection, pinned or preferred path
* [x] Textual hop predicate [sequence](https://scion.docs.anapaya.net/en/latest/PathPolicy.html#sequence) (e.g. for CLI)
* [x] PolicyChain (sequential policy application, And operation)
* [ ] Union / Or operation
* [ ] Sort by path metadata:
  * [x] Lowest latency
  * [ ] Lowest latency enhanced by geo data (fill/correct latency with geo distance divided by speed of light)
  * [x] Highest bandwidth
  * [x] Fewest (total) hops / prefer direct links
* [ ] Filter by path metadata (same list as above but filter with range of allowed values instead of sorting)
* [ ] Textual policy format, e.g. for application configuration (see https://scion.docs.anapaya.net/en/latest/PathPolicy.html)

#### Selector
A path selector is a stateful controller associated with a connection/socket. It receives the paths filtered by the Policy as an input. For each packet sent, the selector choses the path.
The default selector keeps using the first chosen path unless SCMP path down notifications are encountered, in which case it will always switch to the next alive path.
Custom selectors implement e.g. active path probing, coupling of multiple connections to either use the same path or to use maximally disjoint paths, direct performance feedback from the application, etc.

#### Dialed vs Listening
pan differentiates between dialed and listening sockets. Dialed sockets (for "clients") define the path policy and a selector. The client side of a connection is in control of the path used.
The listening side (for "servers") only replies on the paths last used by each client, by means of a customizable reply path selector. The listening side does not implement any policy, nor does it do anything to keep the paths fresh.
The default reply path selector records a fixed number of paths last uses replies on the path last used by a client. It normally uses the path last used by the client, but does use other recorded paths to try routing around temporarily broken paths.

#### ~~PUDP [experiment]~~
Removed from this branch for now: the idea was to look at including explicit path control information in the packets. This control information allows to send packets redundantly over multiple paths (for racing or redundancy), probing, and some path "negotiation". This experiment included this control data in the UDP payload, which is not great for interoperability.  I now think that defining an End-to-End extension header option to transport this control information might be more appropriate.
See documentation in the now removed files for more details: https://github.com/netsec-ethz/scion-apps/blob/4329bc6a7bc6abc2264ad9fd0a3f6fab3e8c3c36/pkg/pan/pudp.go

-----

Changes in this PR:
* add pan library
  * code related to hostnames is almost identical to appnet, but different types etc. Kept mostly private to allow suitably reorganizing this at some point.
* update pkg/shttp to use pan, now allows to specify paths with a Policy
* update scion-bat to use pan, add new command line options to define path choice
* update _examples/helloworld to use pan
* add new _examples/helloquic, to illustrate and test usage of QUIC with pan
* bugfixes backported to appnet: fix SplitHostPort (bad regex capture group index), extend `addrFromString` to work with addresses without the optional brackets.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/187)


juagargi (commented):
This is a great addition! We should merge it ASAP to let people use it in their applications.
I didn't do a full review, but I took some notes. Could you please check them? (non blockers)

- How to hook a statistics processor?
    - Is a subscriber enough?
    - Could the processor push data into the statistics?
- Use case for colibri, in the future we would need:
    - path type in fingerprints
    - refresher must also create renewals
    - scmp path down detection must also check for colibri scmp errors (which are not there yet)
- Policy and Selector are very similar:
    - Both components filter paths.
    - The PingingSelector partially sorts paths (first one is always the less-than, the rest is unsorted)
    - If they are merged into one component, composition (of them) would be easier.

> Reviewed 1 of 43 files at r2.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-:-MlFCSe8E7bpqjJl6qo4:b-x5qq55)* status: 1 of 48 files reviewed, all discussions resolved / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[pkg/pan/stats.go, line 23 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MlEjtDk4XNerh5-Qxgd:-MlEjtDl6UGVAufjRXIk:b-6838vd) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/3f4db7c2b15d686a0d3c9ac73b431b40694bc8de/pkg/pan/stats.go#L23)):*
> ```Go
> )
> 
> var stats pathStatsDB
> ```

Do we have any scenario where we need the info per connection instead of globally?
___
*[pkg/pan/stats.go, line 221 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MlFFDnSAN5fVHjfHS9O:-MlFFDnSAN5fVHjfHS9P:b-c4kx1c) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/3f4db7c2b15d686a0d3c9ac73b431b40694bc8de/pkg/pan/stats.go#L221)):*
> ```Go
> func (s *pathStatsDB) NotifyPathDown(pf PathFingerprint, pi PathInterface) {
> 	s.recordPathDown(pf, pi)
> 	for _, subscriber := range s.subscribers {
> ```

What about running the loop and callbacks in a separate go-routine?

matzf (commented):
> How to hook a statistics processor? 

There is nothing like this yet, and I don't have a clear idea on what the requirements for this would even be. The statistics are really bare bones at this point, and for this reason I've avoided exporting them from the package API for this version.

> Use case for colibri, in the future we would need

Ok, makes sense.


> Policy and Selector are very similar: 

They may be similar on some level of abstraction, but I see them quite distinct; the Policy expresses the user's general preferences and rules. A policy is state-less, "long-lived" and is generically applicable to all destinations (but of course it can contain different rules for different sets of destinations).
The Selector, on the other hand, is stateful, tied to the life-time of a connection and only cares about the specific destination.

The idea is that the Policy filters and sorts paths based on the "static" path information. The Selector then takes into account "dynamic" information (dead paths, latency probes, etc) to select the "dynamically" best choice among the "statically" ranked paths. Perhaps, this makes more sense when I mention that the PingingSelector currently lacks any tuning parameters that weigh the "static" ranking against the "dynamic" information -- how much lower latency does rank `i+1` need to have so we favor it over rank `i` path? I've left this out until now, as it just seems very tricky to parametrize right (and who would ever pick non-default values anyway?), but I should add a TODO for this.

I think I understand what you mean by the composability argument. Originally, this was in fact implemented as you suggest. The following observations made me change this to the current design with separate Policy/Selector concepts:
- The policy result can be cached until the paths are refreshed. If we directly compose Selectors which potentially use "dynamic" information, we'd need additional quirks to avoid re-evaluating the full chain on every packet. Note that some of the ordering operations for the static information are rather expensive.
- Applying a Policy always makes sense, regardless of which Selector is used.
- The default use case, as I imagine it, is to only specify a custom policy and use the DefaultSelector. This gets more awkward if the policy _is_ (part of) the selector.
- Dynamically changing the policy of a connection (which is an important feature) appears to easier if the connection object contains both the policy and the selector separately.

With that, I'd prefer to keep these two concepts separate for this initial version. However, I'm very open if you have a specific suggestion on how this could be improved.

> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-:-Mlo9JGJ57Q673z0R8Gv:blnyhsy)* status: 1 of 48 files reviewed, 1 unresolved discussion / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[pkg/pan/stats.go, line 23 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MlEjtDk4XNerh5-Qxgd:-Mlo-OBQ7BpcgyDD5SFc:b-oxme8i) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/3f4db7c2b15d686a0d3c9ac73b431b40694bc8de/pkg/pan/stats.go#L23)):*
<details><summary><i>Previously, juagargi (Juan A. Garcia Pardo) wrote…</i></summary><blockquote>

Do we have any scenario where we need the info per connection instead of globally?
</blockquote></details>

Good point. The advantages of having this per connection would seem to be: general advantages of less global state (e.g. testability, compatibility of different versions of the library, etc) and information leakage / privacy concerns. The privacy concerns would matter e.g. in a web browser setting where we might want to isolate certain windows from each other.
What would make most sense, I think, would be to allow create multiple instances of the whole global state of the library (e.g. the connection to the sciond, the background runners, and this stats DB) and pass this (in some form) to all the API invocations. This would be quite clean, but I believe it would make the API less convenient to use in the default case, and/or more work to implement correctly. As far as I can tell, at least all the relevant globals are hidden from the API, so implementing this at a later point should be doable.
___
*[pkg/pan/stats.go, line 221 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MlFFDnSAN5fVHjfHS9O:-Mlo-QmF1EHX5Y70qRfA:bh90wl7) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/3f4db7c2b15d686a0d3c9ac73b431b40694bc8de/pkg/pan/stats.go#L221)):*
<details><summary><i>Previously, juagargi (Juan A. Garcia Pardo) wrote…</i></summary><blockquote>

What about running the loop and callbacks in a separate go-routine?
</blockquote></details>

Good point, will do.

marcfrei:
___
*[pkg/pan/udp_dial.go, line 36 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnQsFEW-sJSY2wT8uqk:-MnQsFEW-sJSY2wT8uql:bp7wb8s) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/udp_dial.go#L36)):*
> ```Go
> }
> 
> func DialUDP(ctx context.Context, local *net.UDPAddr, remote UDPAddr,
> ```

Add doc comment?

marcfrei:
___
*[pkg/pan/udp_dial.go, line 30 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnQweLx3Anm9QB0vH-Z:-MnQweLx3Anm9QB0vH-_:b-rizzt5) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/udp_dial.go#L30)):*
> ```Go
> 	// WritePath writes a message to the remote address via the given path.
> 	// This bypasses the path policy and selector used for Write.
> 	WritePath(path *Path, b []byte) (int, error)
> ```

How about WriteToPath

marcfrei:
___
*[pkg/pan/udp_dial.go, line 33 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnQx5K8BKh7s6W-3-sW:-MnQx5K8BKh7s6W-3-sX:bppmdyt) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/udp_dial.go#L33)):*
> ```Go
> 	// ReadPath reads a message and returns the (return-)path via which the
> 	// message was received.
> 	ReadPath(b []byte) (int, *Path, error)
> ```

ReadFromPath?

marcfrei (commented):
> Reviewed 1 of 34 files at r1, 16 of 43 files at r2, 31 of 36 files at r3, 3 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/187)* status: 51 of 76 files reviewed, 22 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[_examples/helloquic/helloquic.go, line 1 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnUxqdH2S3gi30XQYsN:-MnUxqdH2S3gi30XQYsO:b-3qg145) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/_examples/helloquic/helloquic.go#L1)):*
> ```Go
> // Copyright 2021 ETH Zurich
> ```

For consistency, add simple README.md and mention the example in the global README in section `## _examples`
___
*[_examples/helloworld/helloworld.go, line 82 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnUnsdXB4AsxoLNVZ_L:-MnUnsdXB4AsxoLNVZ_M:b-u4vloh) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/_examples/helloworld/helloworld.go#L82)):*
> ```Go
> 	defer conn.Close()
> 
> 	for i := 0; i < 1; i++ {
> ```

Maybe introduce a variable (and a command-line option) for the number of messages to send? 
___
*[netcat/udp.go, line 47 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnUqTl2BG_CHjeTcoMt:-MnUqTl2BG_CHjeTcoMu:b-9n2tom) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/netcat/udp.go#L47)):*
> ```Go
> 
> func (conn *udpListenConn) Write(b []byte) (int, error) {
> 	return conn.write(b)
> ```

Take the mutex here too, for consistency?
___
*[pkg/pan/raw.go, line 57 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnS4qlZ2lwDqbPYqoes:-MnS4qlZ2lwDqbPYqoet:bf9r8gk) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/raw.go#L57)):*
> ```Go
> 	raw         snet.PacketConn
> 	readMutex   sync.Mutex
> 	readBuffer  snet.Bytes
> ```

`readBuffer` is always == nil, right?
___
*[pkg/pan/raw.go, line 59 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnS52062GEnM6OClU8E:-MnS52062GEnM6OClU8F:bv7yds7) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/raw.go#L59)):*
> ```Go
> 	readBuffer  snet.Bytes
> 	writeMutex  sync.Mutex
> 	writeBuffer snet.Bytes
> ```

`writeBuffer` is always == nil, right?
___
*[pkg/pan/raw.go, line 165 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnR885R70DY1QypR1tG:-MnR885R70DY1QypR1tH:b-cb60i1) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/raw.go#L165)):*
> ```Go
> 			underlay: &lastHop,
> 		}
> 		n := copy(b, udp.Payload)
> ```

This copy hurts, doesn't it? Could we introduce an additional type that somehow wraps snet.Bytes/snet.Packet and that could be used to offer more efficient read operations?
___
*[pkg/pan/udp_dial.go, line 76 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnQzjzc1LeMmoMySjwa:-MnQzjzc1LeMmoMySjwb:bnx2axe) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/udp_dial.go#L76)):*
> ```Go
> 	remote     UDPAddr
> 	subscriber *pathRefreshSubscriber
> 	Selector   Selector
> ```

Why is `Selector`exported?
___
*[pkg/quicutil/single.go, line 39 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnUysdj9aQLlKfsslAy:-MnUysdj9aQLlKfsslAz:b-muxi7r) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/quicutil/single.go#L39)):*
> ```
> unidirectional
> ```
___
*[pkg/quicutil/single.go, line 42 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnUzPYlBYFZXsVzWo4H:-MnUzPYlBYFZXsVzWo4I:b4zf92t) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/quicutil/single.go#L42)):*
> ```
> unidirectional
> ```

And replace tabs in comment in this and the next line with spaces.
___
*[ssh/client/ssh/selector.go, line 66 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnUtmvv4Oa9oQDBicED:-MnUtmvv4Oa9oQDBicEE:b-yg3iwy) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/ssh/client/ssh/selector.go#L66)):*
> ```Go
> 	}
> 	p := s.paths[s.current]
> 	s.current += 1
> ```

`% len(s.paths)`?

benthor (commented):
> Reviewed 1 of 43 files at r2.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/187)* status: 52 of 76 files reviewed, 23 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[pkg/pan/selector.go, line 37 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnVfRpy1YCCSXCmfLqG:-MnVfRpy1YCCSXCmfLqH:b-xn4k9x) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/selector.go#L37)):*
> ```Go
> // OnPathDown is invoked when SCMP path down notifications are received.
> type Selector interface {
> 	Path() *Path
> ```

Could this be `Path(UDPAddr)`, to match`SetPath(UDPAddr, []*Path)`?

benthor:
(Turned my above suggestion into a PR matzf/scion-apps#1)

marcfrei (commented):
> Reviewed 3 of 34 files at r1, 17 of 43 files at r2, 4 of 36 files at r3, 1 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/187)* status: all files reviewed, 35 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

*[a discussion](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnVgVNd61E7ruSZPq9Z:-MnVgVNd61E7ruSZPq9_:b-we3qzc) (no related file):*
TODO: search in all files for occurrences of `appnet` and check whether they are still correct/needed.

---
___
*[pkg/pan/cli.go, line 41 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnXSuYW4wlYBRk4IZ9p:-MnXSuYW4wlYBRk4IZ9q:bcnmy74) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/cli.go#L41)):*
> ```Go
> //
> // The options should be presented to the user as:
> //  -	a flag --interactve
> ```

- space instead of tab after '- '
- "interactve" => "interactive"
___
*[pkg/pan/def.go, line 38 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnXAME7B3EIXEvZrx31:-MnXAME7B3EIXEvZrx32:bk1l0yz) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/def.go#L38)):*
> ```
> 	pathExpiryRefreshLeadTime = ...
> 	pathExpiryPruneLeadTime = ...
> 
> ```
I think we could drop `Expiry` from both names.
___
*[pkg/pan/hosts.go, line 34 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnXBDce3IGXTYBZ2rKK:-MnXBDce3IGXTYBZ2rKL:bp85hum) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/hosts.go#L34)):*
> ```Go
> )
> 
> // ResolveUDPAddrAt parses the address and resolves the hostname.
> ```

resolveUDPAddrAt
___
*[pkg/pan/hosts.go, line 90 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnXBobeEkRpAk58y4ZM:-MnXBobeEkRpAk58y4ZN:bbqj2dr) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/hosts.go#L90)):*
> ```Go
> )
> 
> // ParseSCIONAddr converts an SCION address string to a SCION address.
> ```

parseSCIONAddr
___
*[pkg/pan/path.go, line 146 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnXIZfW9anegg_9a70T:-MnXIZfW9anegg_9a70U:b-di0zlq) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/path.go#L146)):*
> ```Go
> }
> 
> // forwardingPathInfo contains information extracted from a dataplane forwardng path.
> ```

forwarding
___
*[pkg/pan/policy.go, line 95 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnXLNdk4GYt8O28IRd_:-MnXLNdk4GYt8O28IRda:bf5n4cz) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/policy.go#L95)):*
> ```Go
> }
> 
> // Sequence is a Policy filtering paths matching a textual pattern. The sequence pattern is
> ```

"policy"
___
*[pkg/pan/policy.go, line 182 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnXMyZkCd9_T4b6v_1x:-MnXMyZkCd9_T4b6v_1y:b-4x5g1c) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/policy.go#L182)):*
> ```Go
> func (p HighestMTU) Filter(paths []*Path) []*Path {
> 	sort.SliceStable(paths, func(i, j int) bool {
> 		return paths[i].Metadata.MTU < paths[j].Metadata.MTU
> ```

Not sure, shouldn't this be `>` if we want to sort in descending order?
___
*[pkg/pan/quic_listen.go, line 38 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnVfx5j5_S1eQQ7EdCb:-MnVfx5k3gF8SrwFSBYp:b-vb6p6q) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/quic_listen.go#L38)):*
> ```
> ListenQUIC
> ```
___
*[pkg/pan/quic_single.go, line 1 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnVhLJU5ayeJNNpe52O:-MnVhLJU5ayeJNNpe52P:bg3hxbr) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/quic_single.go#L1)):*
> ```Go
> // Copyright 2021 ETH Zurich
> ```

Is this file needed?
___
*[pkg/pan/silence.go, line 44 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnWJoGW6NFuy9WlyOsT:-MnWJoGW6NFuy9WlyOsU:b7c80bk) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/silence.go#L44)):*
> ```Go
> 		log.Default().SetOutput(logSilencerOriginal)
> 		logSilencerOriginal = nil
> 	}
> ```

Maybe assert (panic) that `count` is not negative at this point.
___
*[pkg/pan/stats.go, line 158 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnWIBy98C_J23FhS_k9:-MnWIBy98C_J23FhS_kA:b-waz2j4) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/stats.go#L158)):*
> ```Go
> // Returns true if a does not have any recent down notifications and b does, or
> // (more generally) if all down notifications for a are strictly older
> // than any down notificating for b.
> ```

notification

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/187)* status: 55 of 80 files reviewed, 35 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

*[a discussion](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnVgVNd61E7ruSZPq9Z:-Mn_u2Ur3U1-lP7oH7wq:bb23cpc) (no related file):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

TODO: search in all files for occurrences of `appnet` and check whether they are still correct/needed.
</blockquote></details>

Thanks, will do. I'm planning to remove `appnet` with this PR -- just a few more applications to convert:
- bwtester (this one actually requires some thought and work)
- burster / cbrtester
- camerapp ; will be deleted instead (separate PR #208 )

---
___
*[_examples/helloquic/helloquic.go, line 1 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnUxqdH2S3gi30XQYsN:-MnaNJZE90aq2G8BNr2N:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/_examples/helloquic/helloquic.go#L1)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

For consistency, add simple README.md and mention the example in the global README in section `## _examples`
</blockquote></details>

Done.
___
*[_examples/helloworld/helloworld.go, line 82 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnUnsdXB4AsxoLNVZ_L:-Mn_ziy371W-sig9nVqN:b-fznogt) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/_examples/helloworld/helloworld.go#L82)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

Maybe introduce a variable (and a command-line option) for the number of messages to send? 
</blockquote></details>

Good idea, done.
___
*[netcat/udp.go, line 47 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnUqTl2BG_CHjeTcoMt:-MnaNzPB0GSX5a1QzcUi:b-8ilg9) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/netcat/udp.go#L47)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

Take the mutex here too, for consistency?
</blockquote></details>

No I think this wouldn't work, Read and Write can be (safely) called concurrently. As far as i understand, this is actually used concurrently -- we copy (io.Copy) in both directions, from stdin to the connection and from the connection to stdout.

The mutex protects the requests/responses channels.  I've added a comment to clarify this -- this code could definitely be improved a lot though. I'm not convinced this "UDP Listener" abstraction here even makes much sense...
___
*[pkg/pan/cli.go, line 41 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnXSuYW4wlYBRk4IZ9p:-MnkblGY1wzoDAxLEuUk:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/cli.go#L41)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

- space instead of tab after '- '
- "interactve" => "interactive"
</blockquote></details>

Done.
___
*[pkg/pan/def.go, line 38 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnXAME7B3EIXEvZrx31:-MnkcE6UD57I3ssn3huR:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/def.go#L38)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

> ```
> 	pathExpiryRefreshLeadTime = ...
> 	pathExpiryPruneLeadTime = ...
> 
> ```
I think we could drop `Expiry` from both names.
</blockquote></details>

Done.
___
*[pkg/pan/hosts.go, line 34 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnXBDce3IGXTYBZ2rKK:-MnkcIGeBhbnKkxAh2s3:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/hosts.go#L34)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

resolveUDPAddrAt
</blockquote></details>

Done.
___
*[pkg/pan/hosts.go, line 90 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnXBobeEkRpAk58y4ZM:-MnkcJml9mZJhHH3qovN:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/hosts.go#L90)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

parseSCIONAddr
</blockquote></details>

Done.
___
*[pkg/pan/pan.go, line 37 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MmwbPiq-gUHvEQrOuWi:-MnkcMJn1nQrNJ2rYR4n:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/90a288c89599bbe65486e8e91408a5525faea0eb/pkg/pan/pan.go#L37)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

chooses
</blockquote></details>

Done.
___
*[pkg/pan/pan.go, line 52 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-Mmwbe3d2tD237ac8eP6:-MnkcsG99ttjOIfP47vc:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/90a288c89599bbe65486e8e91408a5525faea0eb/pkg/pan/pan.go#L52)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

> ```
> The default reply path selector records a fixed number of paths last uses
> replies on the path last used by a client.
> ```
I think this sentence got broken in an edit.
</blockquote></details>

Done.
___
*[pkg/pan/pan.go, line 91 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-Mmwes910iwxq1rvwNOr:-MnkcwWA1-oBJWZ9A0Hw:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/90a288c89599bbe65486e8e91408a5525faea0eb/pkg/pan/pan.go#L91)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

choose
</blockquote></details>

Done.
___
*[pkg/pan/path.go, line 146 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnXIZfW9anegg_9a70T:-Mnkd-d9Aa070357S7jN:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/path.go#L146)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

forwarding
</blockquote></details>

Done.
___
*[pkg/pan/policy.go, line 95 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnXLNdk4GYt8O28IRd_:-MnkozXT0fCTsuY_gukg:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/policy.go#L95)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

"policy"
</blockquote></details>

Done.
___
*[pkg/pan/policy.go, line 182 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnXMyZkCd9_T4b6v_1x:-MnkdGKT0p319bAHfS98:b-8ft9vm) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/policy.go#L182)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

Not sure, shouldn't this be `>` if we want to sort in descending order?
</blockquote></details>

:100:
___
*[pkg/pan/quic_dial.go, line 27 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-Mmwy3OSDscJXSQbXFcn:-Mnkp0cp3ah5ZSUyA8KC:b-ujdobp) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/90a288c89599bbe65486e8e91408a5525faea0eb/pkg/pan/quic_dial.go#L27)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

> ```
> // This is needed here because we use quic.Dial, not quic.DialAddr but we want
> // the close-the-socket behaviour of quic.DialAddr.
> ```
Does it help to include this implementation detail here?
</blockquote></details>

Hmm yeah, you're right. Removed.
___
*[pkg/pan/quic_dial.go, line 40 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-Mmx-t7J68VOMLG5IF_0:-MnkdoOo4NjJ7jvtTZA-:b-ratglm) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/90a288c89599bbe65486e8e91408a5525faea0eb/pkg/pan/quic_dial.go#L40)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

Not sure, does it make sense to associate the `Policy` concept also with sessions or should clients call `SetPolicy` directly on the connection. 
</blockquote></details>

Not entirely sure either; I've removed it for now.
___
*[pkg/pan/quic_dial.go, line 60 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MmwwVYF2fa-ImdVrmt-:-MnkeaCtDplbMspiZuW0:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/90a288c89599bbe65486e8e91408a5525faea0eb/pkg/pan/quic_dial.go#L60)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

DialQUIC
</blockquote></details>

Done.
___
*[pkg/pan/quic_dial.go, line 62 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-Mmwxg_63m7jG2cYI-LJ:-Mnkea2F4S9_7FR8IDZe:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/90a288c89599bbe65486e8e91408a5525faea0eb/pkg/pan/quic_dial.go#L62)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

> ```
> raddr
> ```
There is no `raddr`. UDPAddr does not contain a path anymore.
</blockquote></details>

Done.
___
*[pkg/pan/quic_dial.go, line 86 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MmwyXyECjeEtfOZhNYm:-MnkeiB-D8KXP8Ajaa4B:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/90a288c89599bbe65486e8e91408a5525faea0eb/pkg/pan/quic_dial.go#L86)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

And DialQUIC instead of DialAddr
</blockquote></details>

Done.
___
*[pkg/pan/quic_listen.go, line 38 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnVfx5j5_S1eQQ7EdCb:-MnkeoED4xHZim_n6a-l:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/quic_listen.go#L38)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

> ```
> ListenQUIC
> ```

</blockquote></details>

Done.
___
*[pkg/pan/raw.go, line 57 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnS4qlZ2lwDqbPYqoes:-MnkqDK8AaNulpsidDoB:bjrxxj9) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/raw.go#L57)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

`readBuffer` is always == nil, right?
</blockquote></details>

Ooh, you're right, great catch! Allocations galore!
Fixed.
___
*[pkg/pan/raw.go, line 59 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnS52062GEnM6OClU8E:-MnktPOn4s91wteGb8hI:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/raw.go#L59)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

`writeBuffer` is always == nil, right?
</blockquote></details>

Done.
___
*[pkg/pan/raw.go, line 165 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnR885R70DY1QypR1tG:-MnkexPWCD5krVEz2izC:brzogny) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/raw.go#L165)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

This copy hurts, doesn't it? Could we introduce an additional type that somehow wraps snet.Bytes/snet.Packet and that could be used to offer more efficient read operations?
</blockquote></details>

It does hurt a bit, yes. As far as I can tell there is really no way around this though. The io.Reader API lets a caller specify a buffer to fill. The result type for this is only number of bytes written to this user buffer, not a slice -- if it was a slice, we could maybe read directly into the user buffer and then cut the headers off. 
So we _could_ provide our own, non-standard ReadSlice (or whatever) function which could avoid this copy, but in many cases we want/need to integrate with existing functionality in libraries that expect the io.Reader API and so we wouldn't really gain much.

FWIW, snet does the same thing here. In the grand scheme of things, this particular copy does not seem to stand out much (at this point in the end host stack, the packet was already copied at least 3 times from kernel to user space buffers or vice versa...).

Btw. I've also thought about whether `recvmsg`'s scatter/gather API could be used to help in any way to avoid this copy (i.e. to make the kernel copy the Payload part directly to the result buffer). With the varying size of the SCION headers, i doubt this is at all possible though.
___
*[pkg/pan/selector.go, line 37 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnVfRpy1YCCSXCmfLqG:-MnkiBB5Ev7Cgsx7Q9d1:b-miy55k) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/selector.go#L37)):*
<details><summary><i>Previously, benthor wrote…</i></summary><blockquote>

Could this be `Path(UDPAddr)`, to match`SetPath(UDPAddr, []*Path)`?
</blockquote></details>

The idea here is that a selector is used as part of a connected socket, so the remote address never changes -- thus, no need to pass it to `Path`.
Btw, it's a not very clean to provide this remote address as part of `SetPaths` -- I did that, just to reduce the number of  functions on the interface, maybe that's not a great idea. Perhaps it would be better to split this, something like this:
```go
Initialize(remoteAddress UDPAddr, paths []*Path)
UpdatePaths(paths []*Path) 
```
___
*[pkg/pan/silence.go, line 44 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnWJoGW6NFuy9WlyOsT:-Mnkl6gj5lep1AJ0l6bS:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/silence.go#L44)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

Maybe assert (panic) that `count` is not negative at this point.
</blockquote></details>

Done.
___
*[pkg/pan/stats.go, line 158 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnWIBy98C_J23FhS_k9:-MnklBn1Ex4eA1yko1sl:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/stats.go#L158)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

notification
</blockquote></details>

Done.
___
*[pkg/pan/udp_dial.go, line 30 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnQweLx3Anm9QB0vH-Z:-Mnkv-TQ7eJq8Yn4yq89:bchi45u) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/udp_dial.go#L30)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

How about WriteToPath
</blockquote></details>

I'd prefer not to use `WriteTo` as prefix as this already has an existing meaning --  `WriteTo`is used for unconnected datagram sockets and let's the caller specify the destination address. I think this could be confusing, as `WriteToPath` would then exist on both connected and unconnected sockets, but with differing signature (with/without destination address). 
Maybe `WriteWithPath` and `ReadWithPath`? Or `WriteByPath` and `ReadByPath`? Seem a bit forced... I don't know. How strong is your dislike for `WritePath`? :sweat_smile: 
___
*[pkg/pan/udp_dial.go, line 76 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnQzjzc1LeMmoMySjwa:-MnklKxG8xQ_eHVq1iDp:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/udp_dial.go#L76)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

Why is `Selector`exported?
</blockquote></details>

Done.
___
*[pkg/quicutil/single.go, line 39 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnUysdj9aQLlKfsslAy:-MnklfjS1grIq5xwn_14:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/quicutil/single.go#L39)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

> ```
> unidirectional
> ```

</blockquote></details>

Done.
___
*[pkg/quicutil/single.go, line 42 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnUzPYlBYFZXsVzWo4H:-MnklfQfE1Y7yejGsg4R:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/quicutil/single.go#L42)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

> ```
> unidirectional
> ```

And replace tabs in comment in this and the next line with spaces.
</blockquote></details>

Done.
___
*[ssh/client/ssh/selector.go, line 66 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnUtmvv4Oa9oQDBicED:-MnkuinAE3tvt9nmXf3v:b22skwr) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/ssh/client/ssh/selector.go#L66)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

`% len(s.paths)`?
</blockquote></details>

Uh, yes that'd be better, thanks! :sweat_smile: 
___
*[pkg/pan/quic_single.go, line 1 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnVhLJU5ayeJNNpe52O:-MnkeqjICdciWk_pg2t8:bddp37h) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/quic_single.go#L1)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

Is this file needed?
</blockquote></details>

Nope, done.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/187)* status: 54 of 80 files reviewed, 35 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[pkg/pan/udp_dial.go, line 36 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnQsFEW-sJSY2wT8uqk:-MnlhkMN-94phnYRQ4qp:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/udp_dial.go#L36)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

Add doc comment?
</blockquote></details>

Done.

marcfrei (approved):
:lgtm:

> Reviewed 1 of 4 files at r5, 1 of 2 files at r6, 6 of 18 files at r7, 64 of 64 files at r10.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-:-Mpq5gyS2z8ZLh3SXfb4:bnfp4nl)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[pkg/pan/silence.go, line 28 at r10](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-Mpq2qJcDU4AcxsNvZT3:-Mpq2qJcDU4AcxsNvZT4:b-kuzf8h) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/11b6299d2e7212690a9be52ce7fcfc54c80a3a60/pkg/pan/silence.go#L28)):*
> ```Go
> // silenceLog redirects the log.Default writer to a black hole.
> // It can be reenabled by calling unsilenceLog.
> // These functions can safely be called from multiple goroutines concurrently;
> ```

What about the following example scenario?
- goroutine A executes `unsilenceLog`, evaluates `count == 0` to `true`, and gets preempted
- while goroutine A is passivated, goroutine B executes and completes `silenceLog`.
- goroutine A gets scheduled again and installs `logSilencerOriginal`. 
___
*[pkg/pan/udp_dial.go, line 30 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnQweLx3Anm9QB0vH-Z:-Mpq-Mq-6apsikYLOZi_:b-vnrtw4) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/udp_dial.go#L30)):*

I'd prefer not to use `WriteTo` as prefix as this already has an existing meaning --  `WriteTo`is used for unconnected datagram sockets and let's the caller specify the destination address. I think this could be confusing, as `WriteToPath` would then exist on both connected and unconnected sockets, but with differing signature (with/without destination address). 
Maybe `WriteWithPath` and `ReadWithPath`? Or `WriteByPath` and `ReadByPath`? Seem a bit forced... I don't know. How strong is your dislike for `WritePath`? :sweat_smile: 
</blockquote></details>

It's not a strong dislike. Other ideas: `WriteViaPath`. Or just `Writep`? Not sure, so maybe `WritePath` is the best alternative after all.

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/187)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[pkg/pan/silence.go, line 28 at r10](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-Mpq2qJcDU4AcxsNvZT3:-Mpq6cfg4jlyPjSkiqg-:b-tuzh65) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/11b6299d2e7212690a9be52ce7fcfc54c80a3a60/pkg/pan/silence.go#L28)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

What about the following example scenario?
- goroutine A executes `unsilenceLog`, evaluates `count == 0` to `true`, and gets preempted
- while goroutine A is passivated, goroutine B executes and completes `silenceLog`.
- goroutine A gets scheduled again and installs `logSilencerOriginal`. 
</blockquote></details>

You're right of course, thanks! I'll just change this to a mutex.
___
*[pkg/pan/stats.go, line 221 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MlFFDnSAN5fVHjfHS9O:-MpMcORt9g9srXGquIAM:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/3f4db7c2b15d686a0d3c9ac73b431b40694bc8de/pkg/pan/stats.go#L221)):*

Good point, will do.
</blockquote></details>

Done.
___
*[pkg/pan/udp_dial.go, line 30 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-MnQweLx3Anm9QB0vH-Z:-Mpq7WH64lB9eOureW6P:bigyg1) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/1b06992ee7372314e41a467561d933fe77753b28/pkg/pan/udp_dial.go#L30)):*
<details><summary><i>Previously, marcfrei (Marc Frei) wrote…</i></summary><blockquote>

It's not a strong dislike. Other ideas: `WriteViaPath`. Or just `Writep`? Not sure, so maybe `WritePath` is the best alternative after all.
</blockquote></details>

Via sounds good, I like it! 

 > Borrowed from Latin viā (“by the way (of)”), ablative singular of via (“way, road”).  

 Road is close enough to "path", so maybe we can omit the redundancy and just use `WriteVia` etc :)

```go
type Conn interface {
    Write(b []byte) (int, error)
    WriteVia(path *Path, b []byte) (int, error)
    Read([] byte) (int, error)
    ReadVia([] byte) (int, *Path, error)
}

type ReplyConn interface {
   WriteTo(b []byte, addr net.Addr) (int,  error)
   WriteToVia(b []byte, addr net.Addr, path *Path) (int, error)
   ReadFrom(b []byte) (int, net.Addr,  error)
   ReadFromVia(b []byte) (int,  net.Addr, *Path,  error)
}
```

I admit that `WriteToVia` and `ReadFromVia` still seem somewhat forced, but to me it seems good enough.

marcfrei (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-:-MpqBSbvEh6VYJJlvr3E:bnfp4nl)* status: :shipit: complete! 1 of 1 LGTMs obtained

marcfrei (commented):
:lgtm:

> Reviewed 4 of 4 files at r11.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/187#-:-MpzOotS4O1Bt9msOTP9:bnfp4nl)* status: :shipit: complete! 1 of 1 LGTMs obtained

matzf:
Pushed a cleaned up history (to be merged with a merge commit).

- add pkg/pan, pkg/quicutils as new libraries
- convert applications to pan (individual commits):
   - web (pkg/shttp, bat, skip, ...)
   - bwtester
   - ssh
   - netcat
   - sensorapp
- remove appnet, appquic

PR #198, opened 2021-08-13
shttp, skip: use net/http instead of quic-go/http3, support tunneling

matzf:
The "default" HTTP package for HTTP over SCION, shttp, now uses HTTP/1
or HTTP/2 based on the net/http standard library.
This allows our applications to talk to "normal" (i.e. non-HTTP/3)
web servers / clients when suitably proxied/tunneled.
Also, this now supports http://, which is useful for quickly running the
demo applications without nasty tricks or having to hassle with TLS
setup.

The existing library for HTTP/3 over SCION is moved to a separate
package shttp3, overhauled and reduced to the bare necessities.
This will need some additional love to make it useful (and used).
Normally HTTP/3 is using the same port as HTTPS but on UDP instead of
TCP -- we run both over UDP so we'll have to come up with something to
disambiguate.

Implement CONNECT in skip to support HTTPS. Change the way that SCION
hosts are identified, as the `.scion`-TLD approach is incompatible with
HTTPS (wrong name for certificates). Instead, it just looks at the list
of hosts for which a SCION address is configured in `/etc/hosts` or
`/etc/scion/hosts`.

Add a web gateway that acts as a proxy/gateway to allow mirroring some
content on the TCP/IP web into the SCION based web.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/198)


matzf:
Note: this is a breaking change, both in terms of API and protocol. Applications using shttp now are not be compatible with applications built before this change as they talk a different protocol (HTTP/1 or 2 now vs HTTP/3 before).

Note 2: I'm planning to add more API breaking changes soon by changing the shttp libraries to "pan", #187.

FR4NK-W (changes requested):
Neat stuff! :tada: 

> Reviewed 22 of 22 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-:-MhP5TmP5ET-N6P5wQu_:b-re4lx4)* status: all files reviewed, 14 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[_examples/shttp/README.md, line 30 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhP3A01-tUg6CcoldjN:-MhP3A01-tUg6CcoldjO:b-r51nue) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/_examples/shttp/README.md#L30)):*
> ```
>  run `example-shttp-server`:
> ```
Nit:
-> "run the `example-shttp-server`:"
___
*[_examples/shttp/README.md, line 74 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhP3dmU5QDNnagif2Wl:-MhP3dmU5QDNnagif2Wm:b-x3zbt1) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/_examples/shttp/README.md#L74)):*
> ```
> obtain **certificate**
> ```
-> "obtain the **certificate**"
___
*[_examples/shttp/server/main.go, line 85 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOzll0B2-Q9Cq5fSw7:-MhOzll10THGMXE-Z5xh:b-v7zy3x) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/_examples/shttp/server/main.go#L85)):*
> ```
> log.Fatal(shttp.ListenAndServe(":80", handler))
> ```

I think it would be nice if we could have the example be "secure", now that we actually use TLS on the outer connection.
I was hopping this was still updated and we could reuse it:
https://github.com/lucas-clemente/quic-go/tree/v0.10.0/example providing valid keys for `quic.clemente.io`.

Or maybe simply duplicate the two lines for the cert and key flags from `shttp/fileserver/main.go` and have a separate section about using the examples with TLS since you are already providing the setup instruction for the certificate later on for the fileserver?
___
*[bat/bat.go, line 193 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOu4F67fdfI6BbY82A:-MhOu4F67fdfI6BbY82B:b-a5fth0) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/bat/bat.go#L193)):*
> ```
> shttp.MangleSCIONAddrURL(proxy)
> ```

Does this work, does bat support connecting to the proxy via SCION?
What is the use case for also connecting to the proxy over SCION?
___
*[pkg/shttp/transport.go, line 36 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOn2V63xqv1gAiG5Xx:-MhOn2V73S9Bk-8sSUWg:b-soifhl) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/pkg/shttp/transport.go#L36)):*
> ```
> }).DialContext,
> ```

Maybe add some more detailed comment here and for the `DialContext` function that we are basically providing a RoundTripper with overridden DialContext.
Setting DialContext also disables HTTP/2, right?
___
*[pkg/shttp3/transport.go, line 1 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOgy0w2Q7rvksw42xB:-MhOgy0x6djt-RohBaOZ:bgin9l5) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/pkg/shttp3/transport.go#L1)):*
> ```
> 2018
> ```

2021 :slightly_smiling_face: 
___
*[skip/main.go, line 33 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOYCn5387dhfEebtV2:-MhOYCn5387dhfEebtV3:bzcay0e) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/skip/main.go#L33)):*
> ```
> text/template
> ```
Should we use package `html/template` to benefit from `JSEscaper`/autoescaping when expanding the `.pac` template to avoid issues with invalid hostnames?
___
*[skip/main.go, line 130 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOc50b0Po_iazhuIxj:-MhOc50b0Po_iazhuIxk:b-6lz0mh) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/skip/main.go#L130)):*
> ```
> other
> ```
Nit:

I'd call the `other` handler the `default` handler.
___
*[skip/main.go, line 159 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOdgZD96Rl_UsapJaf:-MhOdgZD96Rl_UsapJag:b-3238iy) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/skip/main.go#L159)):*
> ```
> w.WriteHeader(http.StatusOK)
> ```

Is there a reason we reply with the HTTP code 200 before we hijack the connection?
Would it not make things easier for debugging and so on, to do that after the hijack? Or is that because we want to keep the CONNECT semantic where the proxy switches to tunnel mode immediately after the successful response?
___
*[skip/main.go, line 164 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOeVxU9dxz6r8WEGNc:-MhOeVxU9dxz6r8WEGNd:b-spowzf) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/skip/main.go#L164)):*
> ```
> clientConn, _, err := hijacker.Hijack()
> ```

Support for HTTP/2 is mentioned in the PR comment, but the docu for the `Hijacker`interface mentions:
>The default ResponseWriter for HTTP/1.x connections supports Hijacker, but HTTP/2 connections intentionally do not.

Is my docu outdated, or is there an other reason why this still would works for HTTP/2 ?
___
*[skip/main.go, line 230 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOg3Cz9I6KlmNagCjI:-MhOg3Cz9I6KlmNagCjJ:b3ut5mq) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/skip/main.go#L230)):*
> ```
> // parseHostsFile, copied/simplified from pkg/appnet/hostsfile.go
> ```

Could we not simply get the keys from the hostsTable?
Or move these helper functions to `pkg/appnet/hostfile.go` to reduce duplication?
___
*[skip/skip.pac, line 2 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOZw-g2sjoFmkrvFUZ:-MhOZw-g2sjoFmkrvFU_:b-5n17q4) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/skip/skip.pac#L2)):*
> ```
> {{range .SCIONHosts}}  "{{.}}",
> ```

I think this breaks if SCIONHosts is empty, has length 0, so => `{{range .SCIONHosts}}  "{{.}}", {{else}} {{end}}`
___
*[web-gateway/main.go, line 115 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhONSz4850PdvvUy8DQ:-MhONSz4850PdvvUy8DR:b-qgafsu) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/web-gateway/main.go#L115)):*
Nit:

> ```
> // Status code
> ```

TLS has "alert codes"/Messages, https://datatracker.ietf.org/doc/html/rfc8446#appendix-B.2 , but I guess that is not applicable here.

>(arbitrary, this is not HTTP).

So maybe just say that we are (re)using the http status codes 501-504 with similar meaning? Or directly only log the error string, and not a status code, if we don't want to define the status codes more precisely.
___
*[web-gateway/README.md, line 18 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOMbSo6jgynjFyxp-G:-MhOMbSp-0VRdXAZjKhW:b-2ufo94) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/web-gateway/README.md#L18)):*
> ```
> scionlab.org www.scionlab.org www.scion-architecture.net
> ```

I guess that should go into a config file in a future version.
Do subdomains always need to be configured separately, or only if they don't share the same TLS certificate/server?

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/198)* status: 10 of 22 files reviewed, 12 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[_examples/shttp/README.md, line 30 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhP3A01-tUg6CcoldjN:-MhQ77UB77ATEJZxQNO-:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/_examples/shttp/README.md#L30)):*

> ```
>  run `example-shttp-server`:
> ```
Nit:
-> "run the `example-shttp-server`:"
</blockquote></details>

Done.
___
*[_examples/shttp/README.md, line 74 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhP3dmU5QDNnagif2Wl:-MhQ79HU9kynVvQbP30s:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/_examples/shttp/README.md#L74)):*

> ```
> obtain **certificate**
> ```
-> "obtain the **certificate**"
</blockquote></details>

Done.
___
*[_examples/shttp/server/main.go, line 85 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOzll0B2-Q9Cq5fSw7:-MhQ7ElaE58-O_u2teiR:b-hx1f4f) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/_examples/shttp/server/main.go#L85)):*

> ```
> log.Fatal(shttp.ListenAndServe(":80", handler))
> ```

I think it would be nice if we could have the example be "secure", now that we actually use TLS on the outer connection.
I was hopping this was still updated and we could reuse it:
https://github.com/lucas-clemente/quic-go/tree/v0.10.0/example providing valid keys for `quic.clemente.io`.

Or maybe simply duplicate the two lines for the cert and key flags from `shttp/fileserver/main.go` and have a separate section about using the examples with TLS since you are already providing the setup instruction for the certificate later on for the fileserver?
</blockquote></details>

Haha, that's pretty funny. I've added the same optional TLS support here, I guess that's the most straight straight forward approach.
Perhaps we could also obtain a similar valid but useless certificate to include in the repository for testing. That'd make the README a bit shorter too :)
___
*[bat/bat.go, line 193 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOu4F67fdfI6BbY82A:-MhQ9bFhAh1c2k8xTVox:bqytwbh) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/bat/bat.go#L193)):*

> ```
> shttp.MangleSCIONAddrURL(proxy)
> ```

Does this work, does bat support connecting to the proxy via SCION?
What is the use case for also connecting to the proxy over SCION?
</blockquote></details>

Yes it works. It has also previously worked for named SCION hosts but not for "naked" SCION addresses. I don't have a specific use case, this is just fixing an edge case in existing functionality. 
___
*[pkg/shttp/transport.go, line 36 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOn2V63xqv1gAiG5Xx:-MhXjNi29qVjnv_nBqbY:b-chwexo) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/pkg/shttp/transport.go#L36)):*

> ```
> }).DialContext,
> ```

Maybe add some more detailed comment here and for the `DialContext` function that we are basically providing a RoundTripper with overridden DialContext.
Setting DialContext also disables HTTP/2, right?
</blockquote></details>

I've read through the http docs a bit and noticed that the `net/http.DefaultTransport` sets `DialContext` too, but also sets `ForceAttemptHTTP2`. I've adapted our code to do the same thing. In the process, I've changed the API a bit, replacing the `NewRoundTripper` function with a `DefaultTransport` variable -- this seems a bit more in-line with the organisation in `net/http`.

I've also extended the doc strings a bit, is this roughly what you meant?
___
*[pkg/shttp3/transport.go, line 1 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOgy0w2Q7rvksw42xB:-MhQATAD7GuzF7r-ewwr:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/pkg/shttp3/transport.go#L1)):*

> ```
> 2018
> ```

2021 :slightly_smiling_face: 
</blockquote></details>

Done.
___
*[skip/main.go, line 33 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOYCn5387dhfEebtV2:-MhQIZDr18s5pU9Cvp0l:b-4m5wio) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/skip/main.go#L33)):*

> ```
> text/template
> ```
Should we use package `html/template` to benefit from `JSEscaper`/autoescaping when expanding the `.pac` template to avoid issues with invalid hostnames?
</blockquote></details>

Great catch! I saw that there is a `js` filter that I've added to the appropriate places in the .pac template. The autoescaping seems a bit daunting to me.
___
*[skip/main.go, line 130 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOc50b0Po_iazhuIxj:-MhQAdks1-QxNWPc4Vpj:b-ceipla) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/skip/main.go#L130)):*

> ```
> other
> ```
Nit:

I'd call the `other` handler the `default` handler.
</blockquote></details>

`default` is a keyword. Renamed it to `next`, this is a "convention" used e.g. in gorilla/mux.
___
*[skip/main.go, line 159 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOdgZD96Rl_UsapJaf:-MhQBQdiBghq-dIpVVBH:b59q5l4) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/skip/main.go#L159)):*

> ```
> w.WriteHeader(http.StatusOK)
> ```

Is there a reason we reply with the HTTP code 200 before we hijack the connection?
Would it not make things easier for debugging and so on, to do that after the hijack? Or is that because we want to keep the CONNECT semantic where the proxy switches to tunnel mode immediately after the successful response?
</blockquote></details>

At this point we've established a connection to the remote host, and we're know that things will work. The only thing that still happens before starting the "tunnel" is to do this Hijacking, which is just a library operation to get access to the underlying connection. Once we've Hijacked, the ResponseWriter can no longer write to this connection so writing the Header afterwards would fail. I guess we could hijack the connection and then manually write the header back, but I can't see any advantage here.
___
*[skip/main.go, line 164 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOeVxU9dxz6r8WEGNc:-MhQCp2oAkhqbPOWawlF:ba64xq2) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/skip/main.go#L164)):*

> ```
> clientConn, _, err := hijacker.Hijack()
> ```

Support for HTTP/2 is mentioned in the PR comment, but the docu for the `Hijacker`interface mentions:
>The default ResponseWriter for HTTP/1.x connections supports Hijacker, but HTTP/2 connections intentionally do not.

Is my docu outdated, or is there an other reason why this still would works for HTTP/2 ?
</blockquote></details>

That's a fair point and honestly I've not looked at or tested HTTP/2 at all, I mentioned it in the docs because net/http mentions it.

Anyway, this here will always work because this proxy is not serving with TLS, and so HTTP/2 is disabled. Added a comment to clarify this.
___
*[skip/main.go, line 230 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOg3Cz9I6KlmNagCjI:-MhQDBg4A7TN4nB3OzOQ:bcmjr89) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/skip/main.go#L230)):*

> ```
> // parseHostsFile, copied/simplified from pkg/appnet/hostsfile.go
> ```

Could we not simply get the keys from the hostsTable?
Or move these helper functions to `pkg/appnet/hostfile.go` to reduce duplication?
</blockquote></details>

Because that's not exported and I think that it shouldn't be.
This here is a hack; there is currently no good way to select for which hosts this proxy should be used and for which it shouldn't be, and this here won't make any sense once RAINS/DNS is really used  -- a different mechanism will be needed, either something configurable in the browser UI or some racing/happy eyeballs mechanism here in skip. 
I prefer to not add functionality to the API that is only used for hacks. Duplicating a few dozen lines of code seems more appropriate ;)
___
*[skip/skip.pac, line 2 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOZw-g2sjoFmkrvFUZ:-MhQEYj2C3B1VvGbcOlp:bugqbe6) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/skip/skip.pac#L2)):*

> ```
> {{range .SCIONHosts}}  "{{.}}",
> ```

I think this breaks if SCIONHosts is empty, has length 0, so => `{{range .SCIONHosts}}  "{{.}}", {{else}} {{end}}`
</blockquote></details>

The docs for this `range` statement say: "If the value of the pipeline has length zero, nothing is output;". This seems what we want, and I've just also just tested it it, looks ok. It expands to:
```js
const scionHosts = new Set([
])
```
___
*[web-gateway/main.go, line 115 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhONSz4850PdvvUy8DQ:-MhQKW6N4Q1ExazG_ByU:bd5nm6h) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/web-gateway/main.go#L115)):*
> So maybe just say that we are (re)using the http status codes 501-504 with similar meaning.

That's exactly what i intended to say. I've rephrased it a bit.
The idea was really to spit out something like the "Common Log Format" (as written out by `LoggingHandler`), to keep it somewhat consistent and, potentially, machine readable. And that only has room for a code, not a message.
___
*[web-gateway/README.md, line 18 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOMbSo6jgynjFyxp-G:-MhQL_OY8Bd18KLVWP2v:bdt4z28) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/web-gateway/README.md#L18)):*

> ```
> scionlab.org www.scionlab.org www.scion-architecture.net
> ```

I guess that should go into a config file in a future version.
Do subdomains always need to be configured separately, or only if they don't share the same TLS certificate/server?
</blockquote></details>

Sure, we could have config file.

Yes, every subdomain needs to be configured; we use the normal ServeMux to route the requests to individual proxy handlers, and this needs to be set explicitly for every subdomain, AFAIU. We could get rid of this restriction, by doing things a bit more manually.

FR4NK-W (changes requested):
> Reviewed 5 of 7 files at r2, 7 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/198)* status: all files reviewed, 2 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[_examples/shttp/client/main.go, line 39 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhYSNLA21u9Jrc-Zyin:-MhYSNLA21u9Jrc-Zyio:biygkvy) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/3d631b1ebc5c58689a983e882d27cf9c429d05a1/_examples/shttp/client/main.go#L39)):*
> ```
> // Create a standard client with our custom RoundTripper
> ```

-> "// Create a client with our Transport/Dialer:" to match the README
___
*[pkg/shttp/transport.go, line 36 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOn2V63xqv1gAiG5Xx:-MhYSoEB7lJMeQ5d3zf1:b-gjt0x1) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/pkg/shttp/transport.go#L36)):*

I've read through the http docs a bit and noticed that the `net/http.DefaultTransport` sets `DialContext` too, but also sets `ForceAttemptHTTP2`. I've adapted our code to do the same thing. In the process, I've changed the API a bit, replacing the `NewRoundTripper` function with a `DefaultTransport` variable -- this seems a bit more in-line with the organisation in `net/http`.

I've also extended the doc strings a bit, is this roughly what you meant?
</blockquote></details>

Yes :100: 
___
*[skip/main.go, line 159 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOdgZD96Rl_UsapJaf:-MhYTJZE1wjgYp5NHzG2:b32w0x) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/skip/main.go#L159)):*

At this point we've established a connection to the remote host, and we're know that things will work. The only thing that still happens before starting the "tunnel" is to do this Hijacking, which is just a library operation to get access to the underlying connection. Once we've Hijacked, the ResponseWriter can no longer write to this connection so writing the Header afterwards would fail. I guess we could hijack the connection and then manually write the header back, but I can't see any advantage here.
</blockquote></details>

OK
___
*[skip/main.go, line 230 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOg3Cz9I6KlmNagCjI:-MhYUm5O93ID5_zdVI8p:b3bug6x) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/skip/main.go#L230)):*

Because that's not exported and I think that it shouldn't be.
This here is a hack; there is currently no good way to select for which hosts this proxy should be used and for which it shouldn't be, and this here won't make any sense once RAINS/DNS is really used  -- a different mechanism will be needed, either something configurable in the browser UI or some racing/happy eyeballs mechanism here in skip. 
I prefer to not add functionality to the API that is only used for hacks. Duplicating a few dozen lines of code seems more appropriate ;)
</blockquote></details>

Fair point
___
*[skip/skip.pac, line 2 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOZw-g2sjoFmkrvFUZ:-MhYVNPq7rRH0D2-Gx6X:b-xmdlhm) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/skip/skip.pac#L2)):*

The docs for this `range` statement say: "If the value of the pipeline has length zero, nothing is output;". This seems what we want, and I've just also just tested it it, looks ok. It expands to:
```js
const scionHosts = new Set([
])
```
</blockquote></details>

True, my bad.
___
*[web-gateway/README.md, line 18 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOMbSo6jgynjFyxp-G:-MhYZP2YEnbW-9fvMCZw:bw5bqv3) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/web-gateway/README.md#L18)):*

Sure, we could have config file.

Yes, every subdomain needs to be configured; we use the normal ServeMux to route the requests to individual proxy handlers, and this needs to be set explicitly for every subdomain, AFAIU. We could get rid of this restriction, by doing things a bit more manually. 
</blockquote></details>

I think it's fine to configure subdomains explicitly as long as we mention it here.

FR4NK-W (approved):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/198)* status: all files reviewed, 2 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/198)* status: 20 of 22 files reviewed, 1 unresolved discussion / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[_examples/shttp/client/main.go, line 39 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhYSNLA21u9Jrc-Zyin:-MhYqs2e3WAzMbHU9mQ2:b-el33r5) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/3d631b1ebc5c58689a983e882d27cf9c429d05a1/_examples/shttp/client/main.go#L39)):*

> ```
> // Create a standard client with our custom RoundTripper
> ```

-> "// Create a client with our Transport/Dialer:" to match the README
</blockquote></details>

Done. +1
___
*[web-gateway/README.md, line 18 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-MhOMbSo6jgynjFyxp-G:-MhYrO56E-aJtZcJcBQC:b-fzgs8y) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/f41b5ec824a8132061ec78136ade459d91cb42e4/web-gateway/README.md#L18)):*

I think it's fine to configure subdomains explicitly as long as we mention it here.
</blockquote></details>

Added a note. :+1:

FR4NK-W (approved):
:lgtm:

> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/198#-:-MjYUR8C1TO8lrpViNt3:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marcfrei)

PR #202, opened 2021-10-05
appquic: silence quic-go warnings

### LINKED ISSUE #201: quic-go library prints a warning about receive buffer size
Thanks @matzf for the changes in 08ad6182351e3fa33f67869a18818f0022cf70f1, they are actually very helpful for implementing the ISL course.

I noticed that when I use the `shttp` package, the `quic-go` dependency is complaining about not being able to set the receive buffer size.

This is not critical, as the HTTP operations still work, but I'm trying to avoid irritating users that would see this message.

# Steps to reproduce

```
make example-shttp-client
./bin/example-shttp-client -s 17-ffaa:0:1107,1.2.3.4
```
(no server needed to be running there, it's just a dummy address)

# Output

```
2021/10/01 15:30:07 connection doesn't allow setting of receive buffer size. Not a *net.UDPConn?. See https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size for details.
```

# Additional information
 
The problem seems to be [at this call](https://github.com/lucas-clemente/quic-go/blob/master/packet_handler_map.go#L85).

matzf:
quic-go wants to set the read buffer size for the Conn, but snet and the
dispatcher setup does not support this.
Faking the operations to set the read buffer size does not work, as
quic-go also checks directly with a system call whether the operation
was successful.
As a hacky workaround, we simply silence the logger during the quic.Dial
and quic.Listen calls.

Closes #201

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/202)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ jobs:
 
   integration:
     machine:
-      image: ubuntu-2004:202101-01
+      image: ubuntu-2004:202107-02
 
     steps:
       - checkout
@@ -92,9 +92,6 @@ jobs:
       - run:
           name: Integration tests
           command: |
-            # XXX: hack, don't build skip here, needs go-1.16, currently this runs on go-1.15.
-            # As there are currently no integration tests for scion-skip, we don't need it here.
-            sed -i '/^\s*scion-skip\s*\\$/d' Makefile
             make integration
       - store_artifacts:
           path: /tmp/scion-apps-integration/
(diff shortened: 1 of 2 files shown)


juagargi (approved):
:lgtm: just needs go 1.16 for `log.Default` (or use `SetOutput`)

> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/202#-:-MlF9w_H8gIOpezwz9kQ:bfdczjq)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #205, opened 2021-10-14
bwtester: tests and fixes for parameter parsing

matzf:
Add tests for `parseBwtestParameters`, with small changes to function
interface to allow calling this from the tests.
Fix discrepancy in computation of bandwidth from other parameters which
had previously led to valid command line parameters to be rejected (for
example `1,?,?,1M`).

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/205)

PR #204, opened 2021-10-14
netcat: add -N/-q, cleanup logging

matzf:
Add functionality -N/-q to shutdown connection after stdin is
closed, analogous to the corresponding flags in BSD netcat.

Cleanup logging; use only one logger library, write logs to stderr (not
stdout). Simplify the different log levels, keep only debug or error
messages, and correspondingly only have a -v flag for for debug logs.

Integrate the netcat/modes/ subpackage into the main netcat package to
simplify the structure, and to share helper functions (for logging).

Note: disabled a "NoPanic" check in the integration tests that depends on specific logging messages. This will be fixed in a follow-up.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/204)

PR #206, opened 2021-10-14
integration test cleanup

matzf:
Cleanup and simplify integration tests and test framework.

- collect full stdout/sterr output into a buffer to allow matching on
  complete output. This simplifies the output matchers (in particular,
  `NoPanic`) and allows to write out much more helpful error messages.

  Unfortunately, the os/exec package has some quirks when capturing
  stdout/stderr of grand-child processes, requiring some workarounds.
  This is described in the new helper function `InputPipeScript`, used
  (only) in the sensorapp integration test.
- add `DefaultIAPairs` function which returns shortened list of test
  pairs with reasonable subset of relevant cases.
- simplify table-driven tests with single case entry
- netcat: completely overhaul integration test setup. Needs new -q flag.
- bwtester: provide duration in test parameters to shorten run
- camerapp: change regex to allow 'T' (timeout with re-try) in output
- change `appnet.Listen` to ensure "ReadySignal" is only printed when
  Listen is successful
- remove various quirks with Chdir around that are not/no longer needed
- call `integration.Init` from `TestMain` in integration tests instead of
  during the test functions. Previously, flags were not be parsed
  correctly and repeated invocation with `go test --count` would panic.
- Makefile: run `go test` with `--count=1` to ensure integration tests
  are re-run when test files are unchanged.

Depends on #204 and #205


FR4NK-W (approved):
:lgtm: 

> Reviewed 1 of 3 files at r3, 6 of 6 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/206#-:-MmSJTBk5YnbyzDJ2WK_:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #208, opened 2021-11-05
Remove camerapp

matzf:
Remove camerapp. We've not had a running instance for some time. The
image-retrieval was built as a custom UDP-based ~reliable protocol. With
the libraries we have now, it would definitely make much more sense to
build this as a rather simple webserver instead (or at least, use quic).

Remove camerapp functionality from webapp.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/208)


FR4NK-W (approved):
The related documentation should then also be deprecated:
https://docs.scionlab.org/content/apps/access_camera.html

> Reviewed 20 of 20 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/208#-:-MoJ98F_5inlcBWMAGub:b-empih8)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #213, opened 2021-12-07
lint: enable more linters fix many issues

matzf:
Enable many golangci-linters that seem somewhat useful and don't show
ridiculous numbers of issues out of the gate.

- update golangci-lint to latest 1.43.0
- consistent import ordering enforced by gci
- nilnil and nilerr linters actually found various bugs where a nil
  error was returned after detecting an error.
- fix issues in bwtester and ssh so that exclude-rules can be removed:
  - bwtester: requires removing the dot-import of the bwtestlib. For
    this, renamed it to shorter `bwtest` and renamed various items in it
    to remove the redundant Bwtest from the name.
  - ssh: minimally handle various errors, or explicitly ignore them
    Reply the error state for `WantReply` requests.
- enable lint for examples (and fix all issues)

-----

It's clear that enabling more linters won't magically turn bad code good. But it probably doesn't hurt either, and it did find a few bugs.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/213)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,31 +14,47 @@ linters:
     #- typecheck
     #- unused
     #- varcheck
-    - gofmt
-    - stylecheck # replacement for golint
+    - asciicheck
+    - bidichk
+    - contextcheck
     - dupl
+    - durationcheck
+    - errname
+    - errorlint
+    - exportloopref
+    - gci
+    - gofmt
+    - ifshort
+    - importas
+    - makezero
     - misspell
-
+    - nilerr
+    - nilnil
+    - nolintlint
+    - prealloc
+    - stylecheck
+    - thelper
+    - wastedassign
+    - tenv
 
 issues:
   # XXX exclude some linters where there are too many issues to fix now
   exclude-rules:
-    - path: bwtester/
-      linters:
-        - stylecheck
     - path: webapp/
       linters:
         - stylecheck
         - staticcheck
         - errcheck
-    - path: ssh/
-      linters:
-        - errcheck
-        
-
   max-same-issues: 0
 
+linters-settings:
+  gci:
+    local-prefixes: github.com/netsec-ethz/scion-apps
+  exhaustive:
+    default-signifies-exhaustive: true
+
 run:
+  build-tags: integration
   # Bat is mostly copied third-party code that we don't care to fix
   # Modifications are only (?) in bat/bat.go
   skip-dirs:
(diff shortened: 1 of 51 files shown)


matzf (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/213)* status: all files reviewed, 2 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[Makefile, line 18 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/213#-MqKqJQ-AbvLR2z1VWr4:-MqPhvse59U4SrhqMK1P:b-odm02d) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/6ea16480ecf6599a781f8b76e2caccdbec5f3483/Makefile#L18)):*
<details><summary><i>Previously, juagargi (Juan A. Garcia Pardo) wrote…</i></summary><blockquote>

why not simply rename `_examples` to `examples` ?
</blockquote></details>

Fair point. This appears to be a convention in the golang world. The idea is that the go tooling should really ignore the examples directory, to avoid excess clutter for people `go build`ing or `go install`ing this, or using this as a dependency -- at some point, it was even suggested to make an explicit special in the go tooling to ignore the `examples` directory, but this was not done as the  "workaround" with an underscore name works well enough.

I'm honestly not entirely convinced that this is (still) useful, and quickly googling around a bit, I found that opinions seem to differ.
Anyway, it's simple enough to make it work as desired as it is now, and it's unclear which option is preferable, so I think I'd rather is to keep it unchanged for now.
___
*[Makefile, line 36 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/213#-MqKqOohDscmsTrjreu0:-MqPhx5t47dV63ef3n6o:b-9yy7en) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/6ea16480ecf6599a781f8b76e2caccdbec5f3483/Makefile#L36)):*
<details><summary><i>Previously, juagargi (Juan A. Garcia Pardo) wrote…</i></summary><blockquote>

don't we want to clean the examples?
</blockquote></details>

Right, done.

juagargi (approved):
:lgtm:

> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/213#-:-MqU9WdIEdkpMsbVxvY9:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matzf)

PR #217, opened 2022-01-07
bwtester: fix incompatibility in PRG packet data

matzf:
In #209, we introduced a reusable internal buffer for filling the
packets with PRG data. This buffer was not fully reset for each packet,
resulting in a different output and thus an incompatibility with the
older implementation. Only for the first packet, we would get the
"correct" result, all subsequent packets are rejected.

Fix by resetting buffer on each Fill call.

Looking at this code, I've noticed that the packet data is not actually
filled with PRG data. Only the first 16 bytes, plus a few bytes at the
end, are ever filled with the PRG output. We should just simplify this
anyway.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/217)


marcfrei (approved):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/217)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #223, opened 2022-03-08
Enable geofencing

JordiSubira:
This PR includes the whitelisting geofencing in the SKIP proxy.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/223)

### DIFF
--- a/pkg/pan/policy.go
+++ b/pkg/pan/policy.go
@@ -157,12 +157,12 @@ func (acl *ACL) String() string {
 
 // Filter evaluates the interface ACL and returns the set of paths
 // that match the list
-func (l ACL) Filter(paths []*Path) []*Path {
+func (acl *ACL) Filter(paths []*Path) []*Path {
 	wps := make([]snet.Path, len(paths))
 	for i := range paths {
 		wps[i] = snetPathWrapper{wrapped: paths[i]}
 	}
-	wps = l.entries.Eval(wps)
+	wps = acl.entries.Eval(wps)
 	ps := make([]*Path, len(wps))
 	for i := range wps {
 		ps[i] = wps[i].(snetPathWrapper).wrapped
--- a/pkg/pan/policy.go
+++ b/pkg/pan/policy.go
@@ -142,6 +142,19 @@ func NewACL(list []string) (ACL, error) {
 	return ACL{entries: acl}, nil
 }
 
+func (acl *ACL) UnmarshalJSON(input []byte) error {
+	acl.entries = &pathpol.ACL{}
+	return acl.entries.UnmarshalJSON(input)
+}
+
+func (acl *ACL) String() string {
+	output := ""
+	for _, entry := range acl.entries.Entries {
+		output += entry.String() + ", "
+	}
+	return output
+}
+
 // Filter evaluates the interface ACL and returns the set of paths
 // that match the list
 func (l ACL) Filter(paths []*Path) []*Path {
--- a/skip/main.go
+++ b/skip/main.go
@@ -26,6 +26,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"net"
 	"net/http"
@@ -75,9 +76,11 @@ func main() {
 	if bindAddress.IP != nil {
 		mux.Handle(bindAddress.IP.String()+"/skip.pac", http.HandlerFunc(handleWPAD))
 	}
+	mux.Handle("/scion-host", http.HandlerFunc(handleHostListRequest))
+	mux.Handle("/setISDPolicy", http.HandlerFunc(proxy.setISDPolicy))
 	mux.Handle("/", proxy) // everything else
 
-	handler := interceptConnect(http.HandlerFunc(handleTunneling), mux)
+	handler := interceptConnect(http.HandlerFunc(proxy.handleTunneling), mux)
 
 	server := &http.Server{
 		Addr:    bindAddress.String(),
@@ -103,8 +106,21 @@ func handleWPAD(w http.ResponseWriter, req *http.Request) {
 	_, _ = w.Write(buf.Bytes())
 }
 
+func handleHostListRequest(w http.ResponseWriter, req *http.Request) {
+	buf := &bytes.Buffer{}
+	scionHost := loadHosts()
+	if len(scionHost) > 0 {
+		buf.WriteString(scionHost[0])
+	}
+	for _, s := range scionHost[1:] {
+		buf.WriteString("\n" + s)
+	}
+	w.Write(buf.Bytes())
+}
+
 type proxyHandler struct {
 	transport http.RoundTripper
+	policy    pan.Policy
 }
 
 func (h *proxyHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
@@ -114,6 +130,7 @@ func (h *proxyHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 
 	resp, err := h.transport.RoundTrip(req)
 	if err != nil {
+		log.Default().Println("error round trip", err.Error())
 		http.Error(w, err.Error(), http.StatusServiceUnavailable)
 		return
 	}
@@ -131,6 +148,40 @@ func copyHeader(dst, src http.Header) {
 	}
 }
 
+func (h *proxyHandler) setISDPolicy(w http.ResponseWriter, req *http.Request) {
+	var aclInput []string
+	if req.Method == http.MethodPut {
+		body, err := ioutil.ReadAll(req.Body)
+		if err != nil {
+			fmt.Println(err.Error())
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+		aclStringList := strings.Split(string(body), ",")
+		aclStringList = aclStringList[:len(aclStringList)-1]
+		// Whitelist ases
+		for _, entry := range aclStringList {
+			aclInput = append(aclInput, "+ "+entry)
+		}
+		aclInput = append(aclInput, "-")
+
+		acl, err := pan.NewACL(aclInput)
+		log.Default().Println("set local ISD policy to", aclInput)
+
+		if err != nil {
+			fmt.Println(err.Error())
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+		h.policy = acl
+		shttp.DefaultTransport.DialContext = (&shttp.Dialer{
+			QuicConfig: nil,
+			Policy:     h.policy,
+		}).DialContext
+		w.WriteHeader(http.StatusOK)
+	}
+}
+
 // interceptConnect creates a handler that calls the handler connect for the
 // CONNECT method and otherwise the next handler.
 func interceptConnect(connect, next http.Handler) http.HandlerFunc {
@@ -143,7 +194,7 @@ func interceptConnect(connect, next http.Handler) http.HandlerFunc {
 	}
 }
 
-func handleTunneling(w http.ResponseWriter, req *http.Request) {
+func (h *proxyHandler) handleTunneling(w http.ResponseWriter, req *http.Request) {
 	hostAddr, err := pan.ResolveUDPAddr(req.Host)
 	if err != nil {
 		fmt.Println(err.Error())
@@ -154,7 +205,7 @@ func handleTunneling(w http.ResponseWriter, req *http.Request) {
 		context.Background(),
 		netaddr.IPPort{},
 		hostAddr,
-		nil,
+		h.policy,
 		nil,
 		req.Host,
 		&tls.Config{
@@ -163,10 +214,12 @@ func handleTunneling(w http.ResponseWriter, req *http.Request) {
 		},
 		nil)
 	if err != nil {
+		log.Default().Println("error dialing", err.Error())
 		fmt.Println(err.Error())
 		http.Error(w, err.Error(), http.StatusServiceUnavailable)
 		return
 	}
+	log.Default().Println(session.Conn.RemoteAddr())
 	destConn, err := quicutil.NewSingleStream(session)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusServiceUnavailable)
--- a/skip/main.go
+++ b/skip/main.go
@@ -123,7 +123,7 @@ func handleHostListRequest(w http.ResponseWriter, req *http.Request) {
 	for i := 1; i < len(scionHost); i++ {
 		buf.WriteString("\n" + scionHost[i])
 	}
-	w.Write(buf.Bytes())
+	_, _ = w.Write(buf.Bytes())
 }
 
 type isdPolicyHandler struct {
@@ -144,13 +144,13 @@ func (h *isdPolicyHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	}
 
 	var acl pan.ACL
-	acl.UnmarshalJSON(body)
+	err = acl.UnmarshalJSON(body)
 	if err != nil {
 		fmt.Println("verbose: ", err.Error())
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
-	h.output.SetPolicy(acl)
+	h.output.SetPolicy(&acl)
 	fmt.Println("verbose: ", "ACL policy = ", acl.String())
 	w.WriteHeader(http.StatusOK)
 }
--- a/skip/main.go
+++ b/skip/main.go
@@ -80,8 +80,7 @@ func main() {
 	apiMux := http.NewServeMux()
 	apiMux.HandleFunc("/skip.pac", handleWPAD)
 	apiMux.HandleFunc("/scion-host", handleHostListRequest)
-	//apiMux.Handle("/setISDPolicy", setISDPolicyHandler)
-	apiMux.Handle("/setISDPolicy", http.HandlerFunc(isdPolicyHandler.setISDPolicy))
+	apiMux.Handle("/setISDPolicy", isdPolicyHandler)
 
 	mux.Handle("localhost/", apiMux)
 	if bindAddress.IP != nil {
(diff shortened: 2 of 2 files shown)


matzf (changes requested):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/223)* status: all files reviewed, 8 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[skip/main.go, line 80 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxddAV14mxF4eEHPu1W:-MxddAV14mxF4eEHPu1X:bhui3yr) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/48b1e6423ecb0dc69f10f73705d74742d0b8954c/skip/main.go#L80)):*
> ```Go
> 	}
> 	mux.Handle("/scion-host", http.HandlerFunc(handleHostListRequest))
> 	mux.Handle("/setISDPolicy", http.HandlerFunc(proxy.setISDPolicy))
> ```

I guess these should be also be restricted to `localhost/endpoint` or `bind-address/endpoint`, like `skip.pac` above.
Perhaps these endpoints should be combined into a separate "mux" for the API for clarity. Something like this, maybe?
```go
apiMux := http.NewServeMux()
apiMux.HandleFunc("/skip.pac", handleWPAD)
apiMux.HandleFunc("/scion-host", handleHostListRequest)
apiMux.Handle("/setISDPolicy", setISDPolicyHandler)

mux.Handle("localhost/", apiMux)
if bindAddress.IP != nil {
   mux.Handle(bindAddress.IP.String(), apiMux)
}
```
___
*[skip/main.go, line 115 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxdWqB08L9PO-X8PHYM:-MxdWqB19Qh5ZrCIinMy:bwawagg) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/48b1e6423ecb0dc69f10f73705d74742d0b8954c/skip/main.go#L115)):*
> ```Go
> 		buf.WriteString(scionHost[0])
> 	}
> 	for _, s := range scionHost[1:] {
> ```

Panics if len(scionHost) == 0
___
*[skip/main.go, line 151 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxdYuhNBuD6weaJO-9h:-MxdYuhNBuD6weaJO-9i:bq3hl5o) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/48b1e6423ecb0dc69f10f73705d74742d0b8954c/skip/main.go#L151)):*
> ```Go
> }
> 
> func (h *proxyHandler) setISDPolicy(w http.ResponseWriter, req *http.Request) {
> ```

This shouldn't be a member of `proxyHandler`. Instead, this can be a separate handler type that sets the Policy on an `interface SetPolicy(pan.Policy)}`.

```go
type isdPolicyHandler struct {
   output interface{SetPolicy(pan.Policy)}
}

func (h *isdPolicyHandler) ServeHTTP(...) {
   ...
   h.output.SetPolicy(acl)
}
```

Initialize the `output` thing with the `dialer` (see comment below).
___
*[skip/main.go, line 153 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxdgMkx09W1bGIKiMrZ:-MxdgMkx09W1bGIKiMr_:bo2taj2) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/48b1e6423ecb0dc69f10f73705d74742d0b8954c/skip/main.go#L153)):*
> ```Go
> func (h *proxyHandler) setISDPolicy(w http.ResponseWriter, req *http.Request) {
> 	var aclInput []string
> 	if req.Method == http.MethodPut {
> ```

No response otherwise? Maybe
```go
if req.Method != http.MethodPut {
   http.Error(w, ..., http.StatusMethodNotAllowed)
   return
}
...
```
___
*[skip/main.go, line 154 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxdYYBIBPhBG3gZvD3k:-MxdYYBJEaWZSpBoUw7P:bf5b61u) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/48b1e6423ecb0dc69f10f73705d74742d0b8954c/skip/main.go#L154)):*
> ```Go
> 	var aclInput []string
> 	if req.Method == http.MethodPut {
> 		body, err := ioutil.ReadAll(req.Body)
> ```

Just `io.ReadAll`.
___
*[skip/main.go, line 180 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxdX8-O2LX4GmLTXOHa:-MxdX8-PCr9HWXfpbYyf:b-64vro4) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/48b1e6423ecb0dc69f10f73705d74742d0b8954c/skip/main.go#L180)):*
> ```Go
> 			QuicConfig: nil,
> 			Policy:     h.policy,
> 		}).DialContext
> ```

Ok, so what we want is to call the `SetPolicy` function of the `shttp.Dialer`. In contrast to this code, this will also adapt the policy of active connections.
We don't have access to this `Dialer` on the `DefaultTransport`, but that's no problem because we probably shouldn't modify the default anyway (unfortunately, golang does not allow this to be `const`). Instead, we create a new one, `transport, dialer := shttp.NewTransport(...)` for the initialization of the `proxyHandler`.
___
*[skip/main.go, line 208 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxdZpKF7czelQbxFcSW:-MxdZpKF7czelQbxFcSX:b-c1n32f) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/48b1e6423ecb0dc69f10f73705d74742d0b8954c/skip/main.go#L208)):*
> ```Go
> 		netaddr.IPPort{},
> 		hostAddr,
> 		h.policy,
> ```

Again, this only sets the policy for dialing. This can be better, we can call `SetPolicy` at any time during the connection.

We need to track the active connections and call `SetPolicy` on all of them.
This sounds annoying, but looking at this now, we can just replace this entire code here by re- (or ab-?) using the `shttp.Dialer`! This would make this policy setting really easy, as it then becomes the same thing for this tunnel and the normal proxied http connections.

```go
type tunnelHandler struct {
   dialer *shttp.Dialer
}

func (h *tunnelHandler) ServeHTTP(...) {
  destConn, err := dialer.DialContext(context.Background(), "", req.Host
  if err != nil {
    http.Error(w, err.Error(), http.StatusServiceUnavailable)
      return
    }
   w.WriteHeader(http.StatusOK)
   ....
}
```

___
*[skip/main.go, line 222 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-Mxdad4t9BEenP2Lq1GD:-Mxdad4u1g_nNsUUcp1w:b-fsgtvy) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/48b1e6423ecb0dc69f10f73705d74742d0b8954c/skip/main.go#L222)):*
> ```Go
> 		return
> 	}
> 	log.Default().Println(session.Conn.RemoteAddr())
> ```

Maybe this can also log this in the same pretend-common-log-format like in `web-gateway` , see `web-gateway/main.go`, `logForwardTLS`.

JordiSubira (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/223)* status: all files reviewed, 8 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[skip/main.go, line 80 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxddAV14mxF4eEHPu1W:-Mxjist50SYG9crvh4kZ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/48b1e6423ecb0dc69f10f73705d74742d0b8954c/skip/main.go#L80)):*

I guess these should be also be restricted to `localhost/endpoint` or `bind-address/endpoint`, like `skip.pac` above.
Perhaps these endpoints should be combined into a separate "mux" for the API for clarity. Something like this, maybe?
```go
apiMux := http.NewServeMux()
apiMux.HandleFunc("/skip.pac", handleWPAD)
apiMux.HandleFunc("/scion-host", handleHostListRequest)
apiMux.Handle("/setISDPolicy", setISDPolicyHandler)

mux.Handle("localhost/", apiMux)
if bindAddress.IP != nil {
   mux.Handle(bindAddress.IP.String(), apiMux)
}
```
</blockquote></details>

Done.
___
*[skip/main.go, line 115 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxdWqB08L9PO-X8PHYM:-MxjjSxD9ghbwQam6bJ6:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/48b1e6423ecb0dc69f10f73705d74742d0b8954c/skip/main.go#L115)):*

Panics if len(scionHost) == 0
</blockquote></details>

Done.
___
*[skip/main.go, line 151 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxdYuhNBuD6weaJO-9h:-MxjjghQ1LHdOndA7lQE:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/48b1e6423ecb0dc69f10f73705d74742d0b8954c/skip/main.go#L151)):*

This shouldn't be a member of `proxyHandler`. Instead, this can be a separate handler type that sets the Policy on an `interface SetPolicy(pan.Policy)}`.

```go
type isdPolicyHandler struct {
   output interface{SetPolicy(pan.Policy)}
}

func (h *isdPolicyHandler) ServeHTTP(...) {
   ...
   h.output.SetPolicy(acl)
}
```

Initialize the `output` thing with the `dialer` (see comment below).
</blockquote></details>

Done.
___
*[skip/main.go, line 153 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxdgMkx09W1bGIKiMrZ:-Mxjjtvr-Sf2mbZzDluS:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/48b1e6423ecb0dc69f10f73705d74742d0b8954c/skip/main.go#L153)):*

No response otherwise? Maybe
```go
if req.Method != http.MethodPut {
   http.Error(w, ..., http.StatusMethodNotAllowed)
   return
}
...
```
</blockquote></details>

Done.
___
*[skip/main.go, line 154 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxdYYBIBPhBG3gZvD3k:-Mxjk1ClC5ANg0fp4mws:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/48b1e6423ecb0dc69f10f73705d74742d0b8954c/skip/main.go#L154)):*

Just `io.ReadAll`.
</blockquote></details>

Done.
___
*[skip/main.go, line 180 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxdX8-O2LX4GmLTXOHa:-Mxjk40gBzAQYKz0DoGG:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/48b1e6423ecb0dc69f10f73705d74742d0b8954c/skip/main.go#L180)):*

Ok, so what we want is to call the `SetPolicy` function of the `shttp.Dialer`. In contrast to this code, this will also adapt the policy of active connections.
We don't have access to this `Dialer` on the `DefaultTransport`, but that's no problem because we probably shouldn't modify the default anyway (unfortunately, golang does not allow this to be `const`). Instead, we create a new one, `transport, dialer := shttp.NewTransport(...)` for the initialization of the `proxyHandler`.
</blockquote></details>

Done.
___
*[skip/main.go, line 208 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxdZpKF7czelQbxFcSW:-Mxjk65x1ZQlf54ntNgG:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/48b1e6423ecb0dc69f10f73705d74742d0b8954c/skip/main.go#L208)):*

Again, this only sets the policy for dialing. This can be better, we can call `SetPolicy` at any time during the connection.

We need to track the active connections and call `SetPolicy` on all of them.
This sounds annoying, but looking at this now, we can just replace this entire code here by re- (or ab-?) using the `shttp.Dialer`! This would make this policy setting really easy, as it then becomes the same thing for this tunnel and the normal proxied http connections.

```go
type tunnelHandler struct {
   dialer *shttp.Dialer
}

func (h *tunnelHandler) ServeHTTP(...) {
  destConn, err := dialer.DialContext(context.Background(), "", req.Host
  if err != nil {
    http.Error(w, err.Error(), http.StatusServiceUnavailable)
      return
    }
   w.WriteHeader(http.StatusOK)
   ....
}
```

</blockquote></details>

Done.
___
*[skip/main.go, line 222 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-Mxdad4t9BEenP2Lq1GD:-MxjkGW09Nle8DoY41xh:btivmvv) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/48b1e6423ecb0dc69f10f73705d74742d0b8954c/skip/main.go#L222)):*

Maybe this can also log this in the same pretend-common-log-format like in `web-gateway` , see `web-gateway/main.go`, `logForwardTLS`.
</blockquote></details>

I have replaced everything with `fmt.Println`, otherwise we have a lot of duplicity with the output generated by the wrapped http logging (`LoggingHandler(...)`). I think for debugging purposes this verbose is helpful, at least, at this moment.

Another option might be exploring a custom logger using `CustomLoggingHandler` instead.

matzf (changes requested):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/223)* status: all files reviewed, 2 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[skip/main.go, line 134 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxkIq993tCdvsfCKy08:-MxkIq993tCdvsfCKy09:b-ta3wqk) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/337935ae4df46305b5cf9a41c42b7f1d357efb0b/skip/main.go#L134)):*
> ```Go
> }
> 
> func (h *isdPolicyHandler) setISDPolicy(w http.ResponseWriter, req *http.Request) {
> ```

`ServeHTTP`
___
*[skip/main.go, line 148 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxkLqZM3rxAUHQeyE87:-MxkLqZM3rxAUHQeyE88:b-d04qqs) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/337935ae4df46305b5cf9a41c42b7f1d357efb0b/skip/main.go#L148)):*
> ```Go
> 	}
> 	aclStringList := strings.Split(string(body), ",")
> 	aclStringList = aclStringList[:len(aclStringList)-1]
> ```

Why drop the last entry, is there a trailing comma?

Thinking about this more, this allows to set more fine grained policies than ISD level, so the name of the handler and the endpoint don't seem accurate.
In fact, it would perhaps make sense to accept the actual ACL language here, instead of this restricted language that needs to be fiddled with here. My suggestion would be to accept the full json-encoded ACL (i.e. a json array of strings,  this feeds directly into NewACL). With this, the policy interface in the front end can be extended quite a bit without requiring changes here.

matzf (changes requested):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/223)* status: 0 of 2 files reviewed, 2 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

JordiSubira (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/223)* status: 0 of 2 files reviewed, 2 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[skip/main.go, line 134 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxkIq993tCdvsfCKy08:-MxtPFEu7W_I_VfATqvp:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/337935ae4df46305b5cf9a41c42b7f1d357efb0b/skip/main.go#L134)):*

`ServeHTTP`
</blockquote></details>

Done.
___
*[skip/main.go, line 148 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-MxkLqZM3rxAUHQeyE87:-MxtPFlD45q4fNi2XMC0:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/337935ae4df46305b5cf9a41c42b7f1d357efb0b/skip/main.go#L148)):*

Why drop the last entry, is there a trailing comma?

Thinking about this more, this allows to set more fine grained policies than ISD level, so the name of the handler and the endpoint don't seem accurate.
In fact, it would perhaps make sense to accept the actual ACL language here, instead of this restricted language that needs to be fiddled with here. My suggestion would be to accept the full json-encoded ACL (i.e. a json array of strings,  this feeds directly into NewACL). With this, the policy interface in the front end can be extended quite a bit without requiring changes here.
</blockquote></details>

Done.

matzf (approved):
> Reviewed 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/223)* status: :shipit: complete! all files reviewed, all discussions resolved

matzf (approved):
:lgtm:

Thank you!

> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/223#-:-MxtktTa9nSFUNGSvZfv:b-vdwr1z)* status: :shipit: complete! all files reviewed, all discussions resolved

matzf (approved):
> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/223)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #230, opened 2022-06-08
Resolve SCION address from DNS TXT record

FR4NK-W:
Add new resolver type dnsResolver
Resolve SCION addresses from DNS TXT records

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/230)

### DIFF
--- a/_examples/helloquic/helloquic.go
+++ b/_examples/helloquic/helloquic.go
@@ -106,7 +106,7 @@ func workSession(session quic.Session) error {
 }
 
 func runClient(address string, count int) error {
-	addr, err := pan.ResolveUDPAddr(address)
+	addr, err := pan.ResolveUDPAddr(context.TODO(), address)
 	if err != nil {
 		return err
 	}
--- a/_examples/helloworld/helloworld.go
+++ b/_examples/helloworld/helloworld.go
@@ -75,7 +75,7 @@ func runServer(listen netaddr.IPPort) error {
 }
 
 func runClient(address string, count int) error {
-	addr, err := pan.ResolveUDPAddr(address)
+	addr, err := pan.ResolveUDPAddr(context.TODO(), address)
 	if err != nil {
 		return err
 	}
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/kormat/fmt15 v0.0.0-20181112140556-ee69fecb2656
 	github.com/lucas-clemente/quic-go v0.23.0
 	github.com/mattn/go-sqlite3 v1.14.4
+	github.com/miekg/dns v1.1.27
 	github.com/msteinert/pam v0.0.0-20190215180659-f29b9f28d6f9
 	github.com/netsec-ethz/rains v0.5.0
 	github.com/pelletier/go-toml v1.9.4
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,6 @@ require (
 	github.com/kormat/fmt15 v0.0.0-20181112140556-ee69fecb2656
 	github.com/lucas-clemente/quic-go v0.23.0
 	github.com/mattn/go-sqlite3 v1.14.4
-	github.com/miekg/dns v1.1.27
 	github.com/msteinert/pam v0.0.0-20190215180659-f29b9f28d6f9
 	github.com/netsec-ethz/rains v0.5.0
 	github.com/pelletier/go-toml v1.9.4
--- a/go.sum
+++ b/go.sum
@@ -445,6 +445,8 @@ github.com/mdlayher/raw v0.0.0-20191009151244-50f2db8cc065/go.mod h1:7EpbotpCmVZ
 github.com/microcosm-cc/bluemonday v1.0.1/go.mod h1:hsXNsILzKxV+sX77C5b8FSuKF00vh2OMYv+xgHpAMF4=
 github.com/miekg/dns v1.0.14/go.mod h1:W1PPwlIAgtquWBMBEV9nkV9Cazfe8ScdGz/Lj7v3Nrg=
 github.com/miekg/dns v1.1.26/go.mod h1:bPDLeHnStXmXAq1m/Ch/hvfNHr14JKNPMBo3VZKjuso=
+github.com/miekg/dns v1.1.27 h1:aEH/kqUzUxGJ/UHcEKdJY+ugH6WEzsEBBSPa8zuy1aM=
+github.com/miekg/dns v1.1.27/go.mod h1:KNUDUusw/aVsxyTYZM1oqvCicbwhgbNgztCETuNZ7xM=
 github.com/mitchellh/cli v1.0.0/go.mod h1:hNIlj7HEI86fIcpObd7a0FcrxTWetlwJDGcceTlRvqc=
 github.com/mitchellh/cli v1.1.0/go.mod h1:xcISNoH86gajksDmfB23e/pu+B+GeFRMYmoHXxx3xhI=
 github.com/mitchellh/go-homedir v1.0.0/go.mod h1:SfyaCUpYCn1Vlf4IUYiD9fPX4A5wJrkLzIz1N1q0pr0=
@@ -956,6 +958,7 @@ golang.org/x/tools v0.0.0-20191115202509-3a792d9c32b2/go.mod h1:b+2E5dAYhXwXZwtn
 golang.org/x/tools v0.0.0-20191119224855-298f0cb1881e/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
 golang.org/x/tools v0.0.0-20191125144606-a911d9008d1f/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
 golang.org/x/tools v0.0.0-20191130070609-6e064ea0cf2d/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
+golang.org/x/tools v0.0.0-20191216052735-49a3e744a425/go.mod h1:TB2adYChydJhpapKDTa4BR/hXlZSLoq2Wpct/0txZ28=
 golang.org/x/tools v0.0.0-20191216173652-a0e659d51361/go.mod h1:TB2adYChydJhpapKDTa4BR/hXlZSLoq2Wpct/0txZ28=
 golang.org/x/tools v0.0.0-20191227053925-7b8e75db28f4/go.mod h1:TB2adYChydJhpapKDTa4BR/hXlZSLoq2Wpct/0txZ28=
 golang.org/x/tools v0.0.0-20200117161641-43d50277825c/go.mod h1:TB2adYChydJhpapKDTa4BR/hXlZSLoq2Wpct/0txZ28=
--- a/go.sum
+++ b/go.sum
@@ -445,8 +445,6 @@ github.com/mdlayher/raw v0.0.0-20191009151244-50f2db8cc065/go.mod h1:7EpbotpCmVZ
 github.com/microcosm-cc/bluemonday v1.0.1/go.mod h1:hsXNsILzKxV+sX77C5b8FSuKF00vh2OMYv+xgHpAMF4=
 github.com/miekg/dns v1.0.14/go.mod h1:W1PPwlIAgtquWBMBEV9nkV9Cazfe8ScdGz/Lj7v3Nrg=
 github.com/miekg/dns v1.1.26/go.mod h1:bPDLeHnStXmXAq1m/Ch/hvfNHr14JKNPMBo3VZKjuso=
-github.com/miekg/dns v1.1.27 h1:aEH/kqUzUxGJ/UHcEKdJY+ugH6WEzsEBBSPa8zuy1aM=
-github.com/miekg/dns v1.1.27/go.mod h1:KNUDUusw/aVsxyTYZM1oqvCicbwhgbNgztCETuNZ7xM=
 github.com/mitchellh/cli v1.0.0/go.mod h1:hNIlj7HEI86fIcpObd7a0FcrxTWetlwJDGcceTlRvqc=
 github.com/mitchellh/cli v1.1.0/go.mod h1:xcISNoH86gajksDmfB23e/pu+B+GeFRMYmoHXxx3xhI=
 github.com/mitchellh/go-homedir v1.0.0/go.mod h1:SfyaCUpYCn1Vlf4IUYiD9fPX4A5wJrkLzIz1N1q0pr0=
@@ -958,7 +956,6 @@ golang.org/x/tools v0.0.0-20191115202509-3a792d9c32b2/go.mod h1:b+2E5dAYhXwXZwtn
 golang.org/x/tools v0.0.0-20191119224855-298f0cb1881e/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
 golang.org/x/tools v0.0.0-20191125144606-a911d9008d1f/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
 golang.org/x/tools v0.0.0-20191130070609-6e064ea0cf2d/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
-golang.org/x/tools v0.0.0-20191216052735-49a3e744a425/go.mod h1:TB2adYChydJhpapKDTa4BR/hXlZSLoq2Wpct/0txZ28=
 golang.org/x/tools v0.0.0-20191216173652-a0e659d51361/go.mod h1:TB2adYChydJhpapKDTa4BR/hXlZSLoq2Wpct/0txZ28=
 golang.org/x/tools v0.0.0-20191227053925-7b8e75db28f4/go.mod h1:TB2adYChydJhpapKDTa4BR/hXlZSLoq2Wpct/0txZ28=
 golang.org/x/tools v0.0.0-20200117161641-43d50277825c/go.mod h1:TB2adYChydJhpapKDTa4BR/hXlZSLoq2Wpct/0txZ28=
--- a/netcat/quic.go
+++ b/netcat/quic.go
@@ -65,7 +65,7 @@ func DoListenQUIC(port uint16) (chan io.ReadWriteCloser, error) {
 
 // DoDialQUIC dials with a QUIC socket
 func DoDialQUIC(remote string, policy pan.Policy) (io.ReadWriteCloser, error) {
-	remoteAddr, err := pan.ResolveUDPAddr(remote)
+	remoteAddr, err := pan.ResolveUDPAddr(context.TODO(), remote)
 	if err != nil {
 		return nil, err
 	}
--- a/netcat/udp.go
+++ b/netcat/udp.go
@@ -55,7 +55,7 @@ func (conn *udpListenConn) Close() error {
 
 // DoDialUDP dials with a UDP socket
 func DoDialUDP(remote string, policy pan.Policy) (io.ReadWriteCloser, error) {
-	remoteAddr, err := pan.ResolveUDPAddr(remote)
+	remoteAddr, err := pan.ResolveUDPAddr(context.TODO(), remote)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/pan/dns_txt.go
+++ b/pkg/pan/dns_txt.go
@@ -53,7 +53,7 @@ func queryTXTRecord(ctx context.Context, host string) (addresses []string, err e
 	var errDNSError *net.DNSError
 	if errors.As(err, &errDNSError) {
 		if errDNSError.IsNotFound {
-			return addresses, err
+			return addresses, HostNotFoundError{Host: host}
 		}
 	}
 	if err != nil {
@@ -65,7 +65,7 @@ func queryTXTRecord(ctx context.Context, host string) (addresses []string, err e
 		}
 	}
 	if len(addresses) == 0 {
-		return addresses, err
+		return addresses, HostNotFoundError{Host: host}
 	}
 	return addresses, nil
 }
--- a/pkg/pan/dns_txt.go
+++ b/pkg/pan/dns_txt.go
@@ -1,0 +1,75 @@
+// Copyright 2022 ETH Zurich
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pan
+
+import (
+	"fmt"
+	"net"
+	"strings"
+
+	"github.com/miekg/dns"
+)
+
+// Open DNS resolver by SWITCH
+const switchResolver1 string = "130.59.31.248"
+const switchResolver2 string = "130.59.31.251"
+
+func init() {
+	resolveDnsTxt = &dnsResolver{}
+}
+
+type dnsResolver struct{}
+
+func (d *dnsResolver) Resolve(name string) (scionAddr, error) {
+	addresses, err := queryTXTRecord(name)
+	if err != nil {
+		return scionAddr{}, err
+	}
+	for _, addr := range addresses {
+		scionAddr, err := parseSCIONAddr(addr)
+		if err == nil {
+			return scionAddr, nil
+		}
+	}
+	return scionAddr{}, fmt.Errorf("error parsing TXT SCION address records")
+}
+
+func queryTXTRecord(query string) (address []string, err error) {
+	msg := new(dns.Msg)
+	if !strings.HasSuffix(query,".") {
+		query += "."
+	}
+	msg.SetQuestion(query, dns.TypeTXT)
+	msg.RecursionDesired = true
+	// TODO : add / import logic to use configured local resolver preference
+	resolver := net.UDPAddr{IP: net.ParseIP(switchResolver1), Port: 53}
+	res, err := dns.Exchange(msg, resolver.String())
+	if err != nil {
+		return address, err
+	}
+	for _, ans := range res.Answer {
+		if txtRecords, ok := ans.(*dns.TXT); ok {
+			for _, txt := range txtRecords.Txt {
+				if strings.HasPrefix(txt, "scion=") {
+					address = append(address, strings.TrimPrefix(txt, "scion="))
+				}
+			}
+		}
+	}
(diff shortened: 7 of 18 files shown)


matzf (changes requested):
Thank you!

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-:-N43qv4C7K21gWCRcVDC:b-9755fn)* status: all files reviewed, 6 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[`pkg/pan/dns_txt.go` line 40 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N43pPuR4umCyKhq6XT0:-N43pPuR4umCyKhq6XT1:b-du750z) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/cfd235449b84a0da809b447b4b554e50fb36de78/pkg/pan/dns_txt.go#L40)):*
> ```Go
> 		}
> 	}
> 	return scionAddr{}, fmt.Errorf("error parsing TXT SCION address records")
> ```

Must return `HostNotFoundError` (see the `Resolver` interface).
___
*[`pkg/pan/dns_txt.go` line 47 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N43pkfsCI6qnwOGi69k:-N43pkftCHw0DagDIbxj:b-tjn3xk) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/cfd235449b84a0da809b447b4b554e50fb36de78/pkg/pan/dns_txt.go#L47)):*
> ```Go
> 		query += "."
> 	}
> 	txtRecords, err := net.LookupTXT(query)
> ```

The `net.Resolver.LookupTxt` interface allows to specify a context (instead of just relying on time out). Maybe this would be a good time to introduce a context to the `Resolve` interface (and the `pan.ResolveUDPAddr`). 
___
*[`pkg/pan/dns_txt.go` line 49 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N43nTFm182cOfuw3iWL:-N43nTFm182cOfuw3iWM:bg1yuuc) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/cfd235449b84a0da809b447b4b554e50fb36de78/pkg/pan/dns_txt.go#L49)):*
> ```Go
> 	txtRecords, err := net.LookupTXT(query)
> 	if err != nil {
> 		return address, err
> ```

Check if it's host not found error (type `*net.DNSError`, `IsNotFound == true`). If so, return `HostNotFoundError` (or just an empty address slice and no error).
___
*[`pkg/pan/dns_txt.go` line 57 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N43m5NtBWB_KilfVx1K:-N43m5NtBWB_KilfVx1L:bawu1oq) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/cfd235449b84a0da809b447b4b554e50fb36de78/pkg/pan/dns_txt.go#L57)):*
> ```Go
> 	}
> 	if len(address) == 0 {
> 		return address, fmt.Errorf("no TXT record with SCION address found")
> ```

No special error case needed here.
___
*[`pkg/pan/hosts.go` line 25 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N43lw0_7reyD0rk6Rux:-N43lw0_7reyD0rk6Ruy:b-7kayoi) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/cfd235449b84a0da809b447b4b554e50fb36de78/pkg/pan/hosts.go#L25)):*
> ```Go
> var (
> 	resolveEtcHosts      resolver = &hostsfileResolver{"/etc/hosts"}
> 	resolveDnsTxt        resolver = nil
> ```

We don't need to the `nil` initialized value and the `init` function here. This was only done for rains to allow building without rains with the `norains` build tag.
___
*[`pkg/pan/hosts.go` line 60 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N43qXlV6kpWcOPsKwF_:-N43qXlV6kpWcOPsKwFa:bdo9u9q) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/cfd235449b84a0da809b447b4b554e50fb36de78/pkg/pan/hosts.go#L60)):*
> ```Go
> //  - /etc/hosts
> //  - /etc/scion/hosts
> //  - RAINS, if a server is configured in /etc/scion/rains.cfg.
> ```

Update docs here and at `pan.ResolveUDPAddr`.
___
*[`pkg/pan/hosts.go` line 66 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N43lxNSBhwlSeFLpQ92:-N43lxNSBhwlSeFLpQ93:b-qwl5ii) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/cfd235449b84a0da809b447b4b554e50fb36de78/pkg/pan/hosts.go#L66)):*
> ```Go
> 		resolveEtcHosts,
> 		resolveDnsTxt,
> 		resolveEtcScionHosts,
> ```

The DNS resolver should come after `/etc/scion/hosts`.

FR4NK-W (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/230)* status: all files reviewed, 6 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[`pkg/pan/dns_txt.go` line 40 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N43pPuR4umCyKhq6XT0:-N47QTX23p_hYKgQcDFR:b-nyafj) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/cfd235449b84a0da809b447b4b554e50fb36de78/pkg/pan/dns_txt.go#L40)):*
I read:

>// Returns a HostNotFoundError if the name was not found, but otherwise no error occurred.

I think that definition of HostNotFoundError you had makes sense. Changed the Resolve function to only return it in this case.
___
*[`pkg/pan/dns_txt.go` line 47 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N43pkfsCI6qnwOGi69k:-N47Vfzh4JxPR0Me2Pve:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/cfd235449b84a0da809b447b4b554e50fb36de78/pkg/pan/dns_txt.go#L47)):*

The `net.Resolver.LookupTxt` interface allows to specify a context (instead of just relying on time out). Maybe this would be a good time to introduce a context to the `Resolve` interface (and the `pan.ResolveUDPAddr`). 
</blockquote></details>

Done.
___
*[`pkg/pan/dns_txt.go` line 49 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N43nTFm182cOfuw3iWL:-N47RAmB44MXS6COHydn:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/cfd235449b84a0da809b447b4b554e50fb36de78/pkg/pan/dns_txt.go#L49)):*

Check if it's host not found error (type `*net.DNSError`, `IsNotFound == true`). If so, return `HostNotFoundError` (or just an empty address slice and no error).
</blockquote></details>

Done.
___
*[`pkg/pan/dns_txt.go` line 57 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N43m5NtBWB_KilfVx1K:-N47ROTeAn7K1T13TP3j:b-jdql7l) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/cfd235449b84a0da809b447b4b554e50fb36de78/pkg/pan/dns_txt.go#L57)):*

No special error case needed here.
</blockquote></details>

I think it is cleaner if we don't rely on the caller, made it return a HostNotFoundError.
___
*[`pkg/pan/hosts.go` line 25 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N43lw0_7reyD0rk6Rux:-N47RZwR-xH_0jxLOjEJ:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/cfd235449b84a0da809b447b4b554e50fb36de78/pkg/pan/hosts.go#L25)):*

We don't need to the `nil` initialized value and the `init` function here. This was only done for rains to allow building without rains with the `norains` build tag.
</blockquote></details>

Done.
___
*[`pkg/pan/hosts.go` line 60 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N43qXlV6kpWcOPsKwF_:-N47VHsl7dOhkXkYB7p1:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/cfd235449b84a0da809b447b4b554e50fb36de78/pkg/pan/hosts.go#L60)):*

Update docs here and at `pan.ResolveUDPAddr`.
</blockquote></details>

Done.
___
*[`pkg/pan/hosts.go` line 66 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N43lxNSBhwlSeFLpQ92:-N47VHCwDV20bLJn13TO:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/cfd235449b84a0da809b447b4b554e50fb36de78/pkg/pan/hosts.go#L66)):*

The DNS resolver should come after `/etc/scion/hosts`.
</blockquote></details>

Done.

FR4NK-W (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/230)* status: all files reviewed, 4 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[`pkg/pan/dns_txt.go` line 57 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N43m5NtBWB_KilfVx1K:-N4CPWeAB8QfRFGij0Ra:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/cfd235449b84a0da809b447b4b554e50fb36de78/pkg/pan/dns_txt.go#L57)):*

Ok. Now it's handled twice though. So remove the error handling for this case in the caller and specify that the function either returns at least one address or an error. 
Btw shouldn't it be called `addresses`?
</blockquote></details>

Done.
___
*[`pkg/pan/hosts.go` line 96 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N48EJ2E-8JDbYedCHkD:-N4CPbEb07ZV5xnyRZn9:b-17xr6p) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/7ba54fc9e73878b8507a618b656dd5fab8cfd6c0/pkg/pan/hosts.go#L96)):*

In this case we should at least check if the context is Done before continuing.
</blockquote></details>

In think we can even completely avoid calling resolvers with an expired context.
___
*[`pkg/pan/pan.go` line 130 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N48AoR1-AqDKcKHuVF3:-N4CPgR5BHGD78z9N7vX:b-54lf98) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/7ba54fc9e73878b8507a618b656dd5fab8cfd6c0/pkg/pan/pan.go#L130)):*

We can just add a context argument to `ResolveUDPAddr`. I think at this point it's still easy to change the API like that. We don't need to take a `cancel` function, this is normally handled by the caller.
</blockquote></details>

I agree that we can still easily change the API to have the apps using the library always provide a context, but I think it is more cumbersome to force all callers to provide a context, even if it is only a TODO context, since most should benefit from some default timeout. Documented the default timeout.
Passing the `cancel` function was a mistake, removed that part.
___
*[`pkg/pan/pan.go` line 142 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N47sElD-dskmPAKkvlX:-N4CQQBZD_WWQJ8w-SO5:b-55f07o) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/7ba54fc9e73878b8507a618b656dd5fab8cfd6c0/pkg/pan/pan.go#L142)):*

The individual resolvers should implement the context handling internally, so there is no need for this device here. Just call the resolver and it will stop when the context expires. 
</blockquote></details>

Not all resolvers implement the context handling, this is the only way I see to ensure that the timeout is guaranteed, even for resolvers like RAINS and hostfile that do not handle the context.
___
*[`pkg/pan/rains.go` line 49 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N48D1apASaGPv9YoJ2D:-N4CQl9R-R5jpaq-L3eI:boy0lyh) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/7ba54fc9e73878b8507a618b656dd5fab8cfd6c0/pkg/pan/rains.go#L49)):*

This should handle the context. Ok for me to leave it for now and just add a `TODO` note. Very likely this doesn't work properly anyway.
</blockquote></details>

There might be some confusion here between the RAINS resolution context, and the execution context from package context.
Renamed the RAINS context.
rains.Query is a function from package rains, that could be modified to take a `Context`, but that is not the case for now.

matzf (changes requested):
> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/230)* status: all files reviewed, 3 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[`pkg/pan/hostsfile.go` line 45 at r3](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N4CWS8xAzXOuHt5OkVd:-N4CWS8yEQkMs-34xOib:bu1caqf) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/82ff416dd8106daa5bd5ab3a5ef1b21145bc482d/pkg/pan/hostsfile.go#L45)):*
> ```Go
> func (r *hostsfileResolver) Resolve(ctx context.Context, name string) (scionAddr, error) {
> 	cachedHostsTable.Lock()
> 	defer cachedHostsTable.Unlock()
> ```

This cache table should be a member of (or just integrated into) the `hostsfileResolver` so that it can be specific to the file path. As it is now, it will be pretty broken if both the `/etc/hosts` and `/etc/scion/hosts` contain addresses.
Btw. separate PR for this?
___
*[`pkg/pan/pan.go` line 130 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N48AoR1-AqDKcKHuVF3:-N4CTTMNAtYzxBHn6bmT:bjqntci) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/7ba54fc9e73878b8507a618b656dd5fab8cfd6c0/pkg/pan/pan.go#L130)):*

I agree that we can still easily change the API to have the apps using the library always provide a context, but I think it is more cumbersome to force all callers to provide a context, even if it is only a TODO context, since most should benefit from some default timeout. Documented the default timeout.
Passing the `cancel` function was a mistake, removed that part.
</blockquote></details>

I don't agree, I think it's worse to have two API variants if it's not strictly necessary. Adding a convenience function later on is easier than removing it.
I know that the `context` API is a bit verbose, but it is what it is.  Having to specify a context forces you to either properly implement timeouts and cancellation, or explicitly ignore it. It should be called `context.YOLO()`. 
Callers will have some context around anyway, e.g. for the Dial that will just come directly afterwards.
___
*[`pkg/pan/pan.go` line 142 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N47sElD-dskmPAKkvlX:-N4CUT7j-zBM2jp5x6w7:b-9ezoos) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/7ba54fc9e73878b8507a618b656dd5fab8cfd6c0/pkg/pan/pan.go#L142)):*

Not all resolvers implement the context handling, this is the only way I see to ensure that the timeout is guaranteed, even for resolvers like RAINS and hostfile that do not handle the context.
</blockquote></details>

Yes but they should. As I commented in the RAINS implementation, it should do this, but it's ok if we defer this (because it's probably not working anyway). But you can move this snippet to the rains implementation if you want.
 The host file implementation does not need to do anything really because it will (normally) be instantaneous anyway. Contexts do not require an immediate hard termination when the context is canceled.

matzf (changes requested):
> Reviewed 7 of 13 files at r4, 7 of 7 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/230)* status: all files reviewed, 5 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[`pkg/pan/hosts.go` line 100 at r5](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N4SJgKBBVsITEII-4w8:-N4SJgKBBVsITEII-4w9:b-n2dhcr) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/3545d8a3501915b7818d915eadd02cbce5e6822a/pkg/pan/hosts.go#L100)):*
> ```Go
> 			break
> 		default:
> 		}
> ```

`if ctx.Err() != nil { break }` is shorter and AFAIU does the same thing.
___
*[`pkg/pan/pan.go` line 130 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N48AoR1-AqDKcKHuVF3:-N4CqS66BYuu_dNlN_l1:b-prees5) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/7ba54fc9e73878b8507a618b656dd5fab8cfd6c0/pkg/pan/pan.go#L130)):*

OK, I'll simply add a `context.TODO()` to all callers.

But I still think that this is not how it should be:
In a way this is comparable to the function "LookupHost" from package net, which also uses a default context, unless called directly on the resolver.
</blockquote></details>

Ok.

As far as I understand, many context-unaware APIs like `net.LookupHost` simply predate the `context` package.
___
*[`pkg/pan/pan.go` line 142 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N47sElD-dskmPAKkvlX:-N4SOhXS-jLim-Wu1pqX:b-9s2vf5) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/7ba54fc9e73878b8507a618b656dd5fab8cfd6c0/pkg/pan/pan.go#L142)):*

> because it will (normally) be instantaneous anyway.

I think that's the whole point of providing a context with timeout here anyway, no?
All of those should be fast _normally_ .
</blockquote></details>

Ok.

There is a large practical difference in the latency and reliability of hosts-file reads (fails/blocks almost never) versus DNS requests (fails at least once a week :wink: ). Ideally, cancellation should also be supported by the hostfile resolver, yes, but as the golang file IO doesn't really support context directly, i think it's just not worth doing.
___
*[`pkg/pan/pan.go` line 113 at r5](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N4SRAMn7kenHDLucLrO:-N4SRAMn7kenHDLucLrP:b-ypd2ek) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/3545d8a3501915b7818d915eadd02cbce5e6822a/pkg/pan/pan.go#L113)):*
> ```Go
> // The address can be of the form of a SCION address (i.e. of the form "ISD-AS,[IP]:port")
> // or in the form of "hostname:port".
> // Default timeout is 1 second, use ResolveUDPAddrContext to provide an explicit context.
> ```

Nit: leftover comment
___
*[`pkg/pan/rains.go` line 67 at r5](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N4SL7ix1Ri-4UUGJOBb:-N4SL7ix1Ri-4UUGJOBc:b-lo9hbz) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/3545d8a3501915b7818d915eadd02cbce5e6822a/pkg/pan/rains.go#L67)):*
> ```Go
> }
> 
> func rainsQuery(ctx context.Context, server UDPAddr, hostname string) (addr scionAddr, err error) {
> ```

Or keep this code unchanged and do the context dance in `rainsResolver.Resolve` (or a separate wrapper func).
___
*[`pkg/pan/rains.go` line 105 at r5](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N4SLLQe4QgOo3bg8IWY:-N4SLLQe4QgOo3bg8IWZ:bq8ihfn) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/3545d8a3501915b7818d915eadd02cbce5e6822a/pkg/pan/rains.go#L105)):*
> ```Go
> 	select {
> 	case <-ctx.Done():
> 		return scionAddr{}, context.DeadlineExceeded
> ```

Nit, `return ..., ctx.Err()`.
___
*[`ssh/server/ssh/tunnelchannel.go` line 86 at r4](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N4CqcP4-Lbz5melQ7gF:-N4CqcP4-Lbz5melQ7gG:bushnx1) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/bfc1d1b28cfd015b5b1dd2af4682d847621b1311/ssh/server/ssh/tunnelchannel.go#L86)):*
> ```Go
> 
> 	ctx := context.Background()
> 	remote, err := pan.ResolveUDPAddr(context.TODO(), address)
> ```

`ctx`, it's right there :smile:

matzf (changes requested):
> Reviewed 3 of 3 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/230)* status: all files reviewed, 1 unresolved discussion / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[`.golangci.yml` line 50 at r7](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-N4XJ3ei-hU5te5KU_PD:-N4XJ3ej3Q_vRwe7TFu_:b-2tyoq8) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/ab0cfbbbb7d9eba2687f582092adff6871339643/.golangci.yml#L50)):*
> ```YAML
>     - path: pkg/pan/selector.go
>       linters:
>         - ifshort
> ```

Separate PR. The update for golangci-lint (mentioned in the commit message) is not included anyway, this is not necessary.

Also, rather just fix it or add a `//nolint:ifshort` to the offending line instead of suppressing it for the entire file.

matzf (approved):
> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/230)* status: :shipit: complete! all files reviewed, all discussions resolved

FR4NK-W (commented):
I'm mostly OK with the PR, but at least the new resolver type should have some tests.

> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-:-N4XwlXTBudme4qsThRd:b-9lye5)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @FR4NK-W)

matzf (changes requested):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/230)* status: all files reviewed, 1 unresolved discussion / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[`pkg/pan/dns_txt.go` line 50 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-NBMiTMU8f6Dnk4E-Dzx:-NBMiTMU8f6Dnk4E-Dzy:b2s1a5h) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/7ba54fc9e73878b8507a618b656dd5fab8cfd6c0/pkg/pan/dns_txt.go#L50)):*
> ```Go
> 	}
> 	resolver := net.Resolver{}
> 	txtRecords, err := resolver.LookupHost(ctx, query)
> ```

Wait, shouldn't this be `LookupTXT` instead of `LookupHost`?
`LookupTXT` was used in r1 of this PR, perhaps this was an accident when changing from `net.LookupTXT` to `net.Resolver.LookupTXT`?

FR4NK-W (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/230)* status: 15 of 16 files reviewed, 1 unresolved discussion / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[`pkg/pan/dns_txt.go` line 50 at r2](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-NBMiTMU8f6Dnk4E-Dzx:-NBNHrDR0nW-fBkACsf0:b-s72q30) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/7ba54fc9e73878b8507a618b656dd5fab8cfd6c0/pkg/pan/dns_txt.go#L50)):*

Wait, shouldn't this be `LookupTXT` instead of `LookupHost`?
`LookupTXT` was used in r1 of this PR, perhaps this was an accident when changing from `net.LookupTXT` to `net.Resolver.LookupTXT`?
</blockquote></details>

Good catch, indeed this is a mistake, introduced in commit 7ba54fc9.
Reverted.

FR4NK-W (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/230)* status: 15 of 16 files reviewed, 1 unresolved discussion / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[`pkg/pan/dns_txt.go` line 56 at r5](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-NBNLFUBB3OxKg5kHASh:-NBNUUwt6GwtQysvR3Ui:b-896fix) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/3545d8a3501915b7818d915eadd02cbce5e6822a/pkg/pan/dns_txt.go#L56)):*

The `HostNotFoundError` was correct here and below.
</blockquote></details>

Done.

matzf (approved):
:lgtm:

> Reviewed 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-:-NBWGsGwBHlQ6RYL6bqS:bnfp4nl)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[`pkg/pan/dns_txt_test.go` line 43 at r11](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-NBWIylCDVdupcOi6CYA:-NBWIylCDVdupcOi6CYB:bbss1vz) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/6d5e21d254958875dc987fbb2829ca50ebb7437b/pkg/pan/dns_txt_test.go#L43)):*
> ```Go
> 	}
> 	var m *mockResolver
> 	resolver := &dnsResolver{res: m}
> ```

Nit, this works fine but looks a bit odd; `m` is `nil` here (but `res` is non-nil because it's an interface). `LookupTXT` will be called with a `nil` receiver. 
For clarity, I'd declare `var m mockResolver` and either pass `res: &m`, or change the signature of `LookupTXT` to a value-receiver `func (r mockupReceiver) LookupTXT(...)`.
___
*[`pkg/pan/dns_txt_test.go` line 60 at r11](https://reviewable.io/reviews/netsec-ethz/scion-apps/230#-NBWH4_9BGTGh6ZlyxpH:-NBWH4_A-e-H4a11rOGo:b-n7duup) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/6d5e21d254958875dc987fbb2829ca50ebb7437b/pkg/pan/dns_txt_test.go#L60)):*
> ```Go
> 
> type mockResolver struct {
> 	net.Resolver
> ```

Nit, embedded `net.Resolver` is not needed

FR4NK-W:
@matzf rebased

FR4NK-W (commented):
> Reviewed 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/230)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #231, opened 2022-06-29
Add strict flag to example fileserver

FR4NK-W:
Add flag strict to example file server for setting the Strict-SCION header if desired to all replies,
explicitly signaling (Strict-)SCION support to clients via the response header.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/231)

### DIFF
--- a/_examples/shttp/fileserver/main.go
+++ b/_examples/shttp/fileserver/main.go
@@ -31,7 +31,7 @@ func main() {
 	certFile := flag.String("cert", "", "Path to TLS server certificate for optional https")
 	keyFile := flag.String("key", "", "Path to TLS server key for optional https")
 	strictSCION := *flag.String("strict", "", "Add `Strict-SCION` header "+
-		"with provided property if not present")
+		"with provided property, replaces any existing value for that header")
 	flag.Parse()
 
 	handler := handlers.LoggingHandler(
@@ -39,8 +39,8 @@ func main() {
 		func(h http.Handler) http.HandlerFunc {
 			return func(w http.ResponseWriter, r *http.Request) {
 				if strictSCION != "" {
-					// Set Strict-SCION response header
-					w.Header().Add("Strict-SCION", strictSCION)
+					// Set Strict-SCION response header, overwrites any existing header for that key
+					w.Header().Set("Strict-SCION", strictSCION)
 				}
 				// Serve
 				h.ServeHTTP(w, r)
--- a/_examples/shttp/fileserver/main.go
+++ b/_examples/shttp/fileserver/main.go
@@ -30,11 +30,22 @@ import (
 func main() {
 	certFile := flag.String("cert", "", "Path to TLS server certificate for optional https")
 	keyFile := flag.String("key", "", "Path to TLS server key for optional https")
+	strictSCION := *flag.String("strict", "", "Add `Strict-SCION` header "+
+		"with provided property if not present")
 	flag.Parse()
 
 	handler := handlers.LoggingHandler(
 		os.Stdout,
-		http.FileServer(http.Dir("")),
+		func(h http.Handler) http.HandlerFunc {
+			return func(w http.ResponseWriter, r *http.Request) {
+				if strictSCION != "" {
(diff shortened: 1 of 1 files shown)


JordiSubira (changes requested):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/231)* status: 0 of 1 files reviewed, 1 unresolved discussion / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[`_examples/shttp/fileserver/main.go` line 33 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/231#-N5smHKQDYXR3fPy0Mv6:-N5smHKQDYXR3fPy0Mv7:b-4fsoyf) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/98e71f90aaf5e4130561381274a5513bceb0ce8e/_examples/shttp/fileserver/main.go#L33)):*
> ```Go
> 	certFile := flag.String("cert", "", "Path to TLS server certificate for optional https")
> 	keyFile := flag.String("key", "", "Path to TLS server key for optional https")
> 	strictSCION := *flag.String("strict", "", "Add `Strict-SCION` header "+
> ```

Could we change this by `flag.Bool(...)`, seems more intuitive.

P.S. This inline flag dereference doesn't work for me. Printing out the strictSCION value shows "" and it never enters the if condition.

FR4NK-W (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/231)* status: 0 of 1 files reviewed, 1 unresolved discussion / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[`_examples/shttp/fileserver/main.go` line 33 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/231#-N5smHKQDYXR3fPy0Mv6:-N5sxSwG6HXP5sgcAd8j:b-gt7cer) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/98e71f90aaf5e4130561381274a5513bceb0ce8e/_examples/shttp/fileserver/main.go#L33)):*
Updated the flag.

>Could we change this by flag.Bool(...), seems more intuitive.

The idea is that one can provide the header value to be set in the "Strict-SCION" header, and not be restricted to a true/false value, but potentially also provide some scope, similar to the HSTS https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security header.
The values that clients handle, and their exact semantic still need to be defined.

JordiSubira (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/231)* status: 0 of 1 files reviewed, all discussions resolved / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[`_examples/shttp/fileserver/main.go` line 33 at r1](https://reviewable.io/reviews/netsec-ethz/scion-apps/231#-N5smHKQDYXR3fPy0Mv6:-N5t1bSj4v5aXBz6PWoA:bnsmcya) ([raw file](https://github.com/netsec-ethz/scion-apps/blob/98e71f90aaf5e4130561381274a5513bceb0ce8e/_examples/shttp/fileserver/main.go#L33)):*

Updated the flag.

>Could we change this by flag.Bool(...), seems more intuitive.

The idea is that one can provide the header value to be set in the "Strict-SCION" header, and not be restricted to a true/false value, but potentially also provide some scope, similar to the HSTS https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security header.
The values that clients handle, and their exact semantic still need to be defined.
</blockquote></details>

As discussed offline, this seems a good approach. We can add a brief description in the usage that references a similar usage to HSTS.

JordiSubira (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/231#-:-N5t2MMK4jCTy1rsXn6k:bnfp4nl)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #234, opened 2022-09-07
Add support for DNS name resolution

benthor:
This PR adds support for DNS name resolution for SCION Addresses in TXT records, which take the form `"scion=ISD-AS,IP"`, see e.g., the output of `dig TXT netsec.ethz.ch.`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/234)


matzf:
Hey Thorben, thanks a lot for your contribution. There is the open PR #230 that seems to implement roughly the same functionality and, as far as I remember, is mostly complete (pending only some more test coverage, self-requested by @FR4NK-W). Is there something that your PR adds that was missing from the other one?

benthor:
Hey Matthias, I hadn't been aware of the other open PR. I prefer @FR4NK-W's approach, since his implementation makes do with the standard library, whereas my quick solution uses a third party lib.

My PR additionally adds support for a scion-specific `resolv.conf`, but I now realize that this feature might do more harm than good because it essentially undermines the system-wide resolver config.

Thanks for not being too discouraging about this, but I myself think PR #230 is the way to go. Sorry for the noise :see_no_evil:

PR #247, opened 2023-12-22
prototype scion-socket

amdfxlucas:
this is required by bindings to languages whose networking libraries main abstraction are berkley sockets, not connections as in Go.It is a prerequisite for i.e. P2P applications where you want to act as both client and server on the same address/socket utilizing QUIC implementations that support it i.e rust-quinn.

- - -
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/247)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,12 +5,21 @@
     "version": "0.2.0",
     "configurations": [
         {
-            "name": "Launch Package",
+            "name": "Launch Client",
             "type": "go",
             "request": "launch",
             "mode": "auto",
             "program": "${workspaceFolder}/_examples/helloworld_socket/helloworld_socket.go",
-            "args": ["-remote" ,"19-ffaa:1:1067,127.0.0.1:2222"]
+            "args": ["-remote" ,"1-151,10.151.0.30:4444"]
+
+        },
+        {
+            "name": "Launch Server",
+            "type": "go",
+            "request": "launch",
+            "mode": "auto",
+            "program": "${workspaceFolder}/_examples/helloworld_socket/helloworld_socket.go",
+            "args": ["-listen" ,"10.150.0.30:4444"]
 
         }
     ]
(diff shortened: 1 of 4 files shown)


matzf:
The background for this change seems somewhat confused; the distinction between client and server side is a peculiarity of pan, not of go the "Conn" concept. Arguably, a "net.Conn"/ "net.PacketConn" _is_ a socket, it's just a different name.

The reason why pan has this separate client/server thing is that we don't want to do a "hidden" path request during a "send" call. Because of this, unconnected sockets are restricted to sending reply packets, as this can be done without an explicit path request. Indeed, the functionality to use a socket for multiple remotes is missing. 
The proposed "DefaultCombiSelector" does exactly what we didn't want to do, namely this "hidden" path request. I'm aware that this path request will eventually need to be "implicit" in some shim-layer, but in pan this was intended to be explicit. 
I'm not saying the pan approach is perfect; but if we add hidden path requests, perhaps the interface could be entirely restructured, instead of adding a separate new "socket" type. I'm not very fond of this proposal to create an interface that is basically the amalgamation of the two variants, as this seems to bloat the library interface with different ways to create connections, and make the common, simple case more complicated.

An alternative approach could be to export a function that allows a user to initialize a path selector analogous to how `DialUDP` does it. This would be an "advanced" interface; the user would explicitly need to manage these, and use `ListenConn.WriteToVia` to send packets with an explicit call to `Selector.Path()`. This would allow using the `ListenConn` to also send packets to any "new" destinations.
Effectively, this would be to making the existing `type pathRefreshSubscriber` and `func openPathRefreshSubscriber` public -- not sure what would be a good name and if the interface needs more cleaning. Basic usage could look something like this:
```go
listenConn, err := pan.ListenUDP(ctx, local, nil)
pather, err := pan.NewPathSubscriber(ctx, local, dstUDPAddr, policy, selector)
listenConn.WriteToVia(msg, dstUDPAddr, pather.Path())
...
pather.Close()
listenConn.Close()
```

PR #253, opened 2024-04-04
pan: fix GetPath if the connection is within the same AS

fbuetler:
As the `c.selector` might be `nil` when [the remote AS and the local AS are the same](https://github.com/netsec-ethz/scion-apps/blob/master/pkg/pan/udp_dial.go#L62-L64), we should respect that in the path retrieval.

cc @JordiSubira 

- - -
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/253)

### DIFF
--- a/pkg/pan/udp_dial.go
+++ b/pkg/pan/udp_dial.go
@@ -98,6 +98,9 @@ func (c *dialedConn) LocalAddr() net.Addr {
 }
 
 func (c *dialedConn) GetPath() *Path {
+	if c.selector == nil {
+		return nil
+	}
 	return c.selector.Path()
 }
 


marcfrei (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/253#-:-Nugy5vH-Dj23EeQjdPH:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

marcfrei (approved):
Thanks, Florian!

> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/253#-:-NugyJpy2g6IrFvZ4oxU:beli9h2)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #259, opened 2024-10-04
pan: fix pathPool query path procedure + close underlay UDPConn if QUIC session was not established successfully

JordiSubira:
There was a bug in the procedure for querying paths, instead of querying paths provided the `pathRefreshMinInterval` had passed, it did it the other way round. I made use of this PR to refactor the internal procedure. In addition, I also added closing the underlay UDPConn if QUIC session was not established successfully.

While troubleshooting the library I also realized that the library is a bit messy and it would be great to spend some effort in refactoring and simplifying it as possible. We kept postponing this, partially, because we discussed about merging [scionproto/snet](https://github.com/scionproto/scion/tree/master/pkg/snet) and the pan library.

- - -
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/259)

### DIFF
--- a/pkg/pan/pool.go
+++ b/pkg/pan/pool.go
@@ -67,16 +67,20 @@ func (p *pathPool) unsubscribe(dstIA IA, s pathPoolSubscriber) {
 }
 
 // paths returns paths to dstIA. This _may_ query paths, unless they have recently been queried.
-func (p *pathPool) paths(ctx context.Context, dstIA IA) ([]*Path, error) {
+func (p *pathPool) paths(ctx context.Context, dstIA IA) ([]*Path, bool, error) {
 	p.entriesMutex.RLock()
 	entry, ok := p.entries[dstIA]
 	p.entriesMutex.RUnlock()
 
-	if ok && !shouldRefresh(time.Now(), entry.earliestExpiry, entry.lastQuery) {
-		return append([]*Path{}, entry.paths...), nil
+	if ok && !shouldQuery(time.Now(), entry.earliestExpiry, entry.lastQuery) {
+		return append([]*Path{}, entry.paths...), false, nil
 	}
 
-	return p.queryPaths(ctx, dstIA)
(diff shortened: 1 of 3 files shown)


FR4NK-W (changes requested):
Thank you. It is now also easier to follow what happens :thumbsup: 

But I believe there is a change in behavior (in addition to the fixing of the refresh after/before expiry):
`refresh(dstIA, paths)` l.128 is now called on the subscribers regardless of the paths having been updated or not, and regardless of the timer.

> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/259#-:-O8MsN6c2EEsh4HwPvqp:bgcbaab)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @juagargi and @marcfrei)

JordiSubira (commented):
Thanks François, you're right, if the paths aren't fetched, it isn't necessary to call refresh on every subscriber (although it is an idempotent operation). I return a boolean flag, to indicate whether the paths are fresh, i.e., recently fetched. It isn't the most elegant solution, but in any case, I am not a big fan of how the library looks like right now and some heavy lifting is anyway necessary.

> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/259#-:-O8My_ej9F6ZnW7TUUlp:b1dyawt)* status: 1 of 3 files reviewed, all discussions resolved / -1 of 2 LGTMs obtained / 0 of 2 approvals obtained

FR4NK-W (approved):
:lgtm:

> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/259#-:-O8aAL3zEvfNyxB_stnA:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @juagargi and @marcfrei)

PR #265, opened 2024-12-24
pan: fix ListenQUIC to close underlay connection

JordiSubira:
Not sure why this BUG was still lingering, perhaps the solution wasn't that obvious for some reason. If we should not close the underlay connection in some scenario, I guess we could rethink the ListenQUIC() implementation. Currently, I've just added a wrapper that closes the underlay connection when Close() is called on the quic listener. Error handling could be improved as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/265)

PR #267, opened 2025-01-27
pan: Fix pool expiry time

JordiSubira:
Prereq: merge #265 

If the query to sciond returned an empty set of paths, the entry is updated with the empty set. The `update()` method though sets the `e.eralisetExpiry = maxTime`. This makes that `shouldQuery()` returns always false for that destination. I updated it so that it returns the zero value for  `time.Time`. 

Note that we will still wait for `earliestAllowedRefresh`, which it is currently a constant (10 seconds), this may be undesirable for some use cases, i.e., applications that cannot wait for this 10 seconds cool-down period by the cost of issuing more request to the daemon.

- - -
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/267)

### DIFF
--- a/pkg/pan/pool.go
+++ b/pkg/pan/pool.go
@@ -129,6 +129,9 @@ func (e *pathPoolDst) update(paths []*Path) {
 func (p *pathPool) earliestPathExpiry() time.Time {
 	p.entriesMutex.RLock()
 	defer p.entriesMutex.RUnlock()
+	if len(p.entries) == 0 {
+		return time.Time{}
+	}
 	ret := maxTime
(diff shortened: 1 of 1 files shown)


tzaeschke (approved):
:lgtm: 

> Reviewed 22 of 22 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/267#-:-OIA6gUv6xLaQylu0oHr:bg7igkv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #273, opened 2025-04-10
Fabrid bandwidth tester

rohrerj:
Contains code that allows to use FABRID for the bandwidth tester.
Since FABRID currently only exists in the netsec-ethz/scion repository, but not in the scionproto repository, go redirects the dependencies to the netsec-ethz/scion repository.

- - -
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/273)

### DIFF
--- a/_examples/go.mod
+++ b/_examples/go.mod
@@ -4,6 +4,8 @@ go 1.21.10
 
 toolchain go1.21.11
 
+replace github.com/scionproto/scion => github.com/netsec-ethz/scion v0.6.1-0.20250410094849-bbacd7f940bc
+
 require (
 	github.com/golang/protobuf v1.5.4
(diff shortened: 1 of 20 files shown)


juagargi (changes requested):
We should have some tests for the added FABRID functionality in `pkg/pan`, at least to know if we break it.
Otherwise the changes LGTM.

> Reviewed 19 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/scion-apps/273#-:-ONxn4Yc5ZEij8neTLwk:b7hksp8)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @rohrerj)

